PR #25, opened 2018-01-30
Make on/off switches accessible

### LINKED ISSUE #24: On/Off toggles are not accessible
All of the on/off toggle widgets on the admin pages are not accessible to assistive technology, such as screen readers for the visually impaired.  Please add the appropriate ARIA markup.  Each of these should have a `role="switch"` or `role="checkbox"`, and also use the `aria-checked` attribute.

Please see the following references:
1. [ARIA 1.1 specification of the switch role](https://www.w3.org/TR/wai-aria-1.1/#switch)
2. [ARIA 1.1 Authoring Practice documentation on checkboxes with examples](https://www.w3.org/TR/wai-aria-practices-1.1/#checkbox)

steverep:
This partially fixes #24 by doing the following:
* Makes switch container and labels positioned relative
* Removes left float of labels (this does nothing)
* Turn on display of radios, but position absolutely and hide behind labels (making them available to assistive technology)
* Add/adjust hover and focus pseudo classes for better accessibility

You'll also need to revert the commit that added `role="switch"` and `aria-checked` to fully fix #24.

### DIFF
--- a/litespeed-cache/css/litespeed.css
+++ b/litespeed-cache/css/litespeed.css
@@ -510,6 +510,7 @@
 	background-color: #fff;
 	padding: -1px;
 	display: inline-block;
+	position: relative;
 }
 
 .litespeed-switch input:checked + label {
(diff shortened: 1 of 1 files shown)


hailite:
Thanks a lot.

PR #43, opened 2018-07-21
Show level progress up to level 7

lucasRolff:
image optimization page doesn't show the progress bar after level 5 - so people do not know how far they are from level 6 and 7.


hailite:
Levels over 5 are not in automatic level up system. It based on couple factors.

JMCA2:
@hailite How do you get to levels 6 or 7? I've optimized over 1500 groups and still on 5

PR #53, opened 2018-10-03
Update lazyload.min.js

wllcrre:
fix lazyload not work on iframe tag


hailite:
Please also update other related lazyload files.

PR #54, opened 2018-10-03
Update lazyload.init.js

wllcrre:
fix lazyload didn't work on iframe tag


hailite:
Please understand why we are using `addEventListener` in the previous codes.

PR #57, opened 2018-10-11
Skip count checking when multisite and use primary site configurations is on

stanleybz:
`🐞 In multiple sites second site keeps receiving purge msg on admin pages`

`plugin_upgrade()` will keep run `purge_all()` when use primary site configurations is on because

```count( $default_options ) == count( $this->options )```

always return false.

Adding

```! is_main_site() && ! empty ( $site_options[ self::NETWORK_OPID_USE_PRIMARY ] )```

to check if use primary site configurations are on and not on main_site.


closte:
Well, we received a ticket today about this same issue:

> Every time I log in a SUBSITE, I have the message Purged all caches successfully. Is that normal?

Hope is the same issue.

PR #58, opened 2018-10-16
remove_query_strings() regex set remove until next & sign

stanleybz:
Prevent error causing by `+` sign on the query string

`e.g.: https://example.com/test.js?ver=1.2.3+pro3.4.5`

Remove all the character until next `&` sign


hailite:
This won't fix the issue `https://example.com/test.js?ver=1.2&aa=bb`

PR #61, opened 2018-10-17
update lazyload.js and lazyload.min.js

wllcrre:
update lazyload.js and lazyload.min.js to verlok's lazyload v10.19


hailite:
Wrong indent chars

PR #70, opened 2018-12-05
Check object exists before loop

stanleybz:
- [Topic] Problem with activation – errors 
- Warning: Invalid argument supplied for foreach() in `/usr/local/lsws/wordpress/html/wp-content/plugins/litespeed-cache/admin/litespeed-cache-admin-settings.class.php on line 1062`
- litespeed-crawler-cookies empty


hailite:
We won't fix it this way. We had a `LiteSpeed_Cache_Config::convert_options_to_input()` to convert the options from final format to the previous status in setting form. This way it will lose the cookie simulation values.

PR #74, opened 2018-12-20
List of Mobile User Agents mobilerules bug fixing

stanleybz:
Add space for List of Mobile User Agents to prevent network mobile rule bug when register_activation in multisite network admin


hailite:
1. Plz use empty() in condition to avoid PHP warning.

2. Can you tell why didn't use `addslashes()` directly? Bcos the issue is not about space didn't have slash but slash needs one more slash.

stanleybz:
Noted, will fix and create another PR.

PR #77, opened 2018-12-26
Fix reset all settings button and copied save changes button

stanleybz:
Align left for [reset all settings] and top [save changes button] when RTL language.


hailite:
A better way is to just use body's `rtl` class as parent selector and modify our css to manipulate the button.

PR #78, opened 2019-01-03
Add thirdparty php and update esi class to support addslashes

stanleybz:
Supporting ESI for DIVI Theme builder, add slashes for the comment box under JS variable.


hailite:
1. Please format all places that needs space.
2. To make it more universal, in `load_comment_form_block()`, don’t use `is_js` as param, change it to `addslashes`, values are the chars that need to add slashes.

stanleybz:
1. Ok
2. Already using addslashes, or I missed something?

stanleybz:
Spacing fixed.

PR #81, opened 2019-01-07
Add thirdparty php and update esi class to support js content

stanleybz:
Supporting ESI for DIVI Theme builder, json encode for the comment box under JS variable.


hailite:
Merged. Please get ruikai test and make sure its fixed.
Test includes:
1) before, can reproduce.
2) after, it's fixed.

PR #82, opened 2019-01-10
[Next Gen Gallery] Handle import from media library to gallery and fix for new version

stanleybz:


hailite:
1. `$_POST['gallery_id']` needs space.
2. Please combine `$_GET['gallery_id']` and `$_POST['gallery_id']` to `$_REQUEST`.
3. Simplify changes to `function update_gallery()`. No need to write purge twice in both 179&183. Use ``` if(....) { $gid=$gid->gid };```

PR #93, opened 2019-01-27
PHP unit method casing

MathiasReker:
Review carefully! Might break tests.


hailite:
PHP unit is temporarily terminated development in our dev plan for a while already. Will leave it till we decided to open it again.

PR #97, opened 2019-01-27
Function declaration

MathiasReker:


hailite:
We keep spaces in func between parentheses and beginning/end params.

PR #99, opened 2019-01-27
Improve type casting

MathiasReker:


hailite:
Thanks but we will fix these in refactoring individually.

PR #107, opened 2019-01-27
Standardize increment

MathiasReker:


hailite:
This is a 3rd party class. We would rather keep it as is if no compatibility issues.

PR #127, opened 2019-02-18
Create hook for checking internal url. Add WPML domains to hook.

stanleybz:
Fixing the following issue by default:

https://wordpress.org/support/topic/litespeed-not-working-with-wpml-secondary-domain/
https://wordpress.org/support/topic/wpml-with-subdomains-litespeed-not-aggregating-resources/


hailite:
The reason defined `LITESPEED_FRONTEND_HOST` in `internal()` is the performance, as this function may be called heavily. So directly using filter inside is not recommended. I will improve it after approval. Just be aware of it in future commits.

hailite:
No worry I will write a 3rd party file for it soon.

hailite:
https://github.com/litespeedtech/lscache_wp/commit/3d9aa5738445b1fff996740dba706e80979709a1

PR #131, opened 2019-03-28
Added DocBlock to JS Defer Exclusion Filter

DaveyJake:
Found excessive `Array to string conversion` errors coming from a custom filter that thought the filter parameter was a string.  The errors stopped when the filter parameter was treated as an array.  The DocBlock makes it clear that the `litespeed_optm_js_defer_exc` filter needs an array and not a string.


hailite:
Thanks, we will give this a review and test soon and get back.

hailite:
This PR is updated in https://github.com/litespeedtech/lscache_wp/commit/3ba334cdfbb314b0cdb613d5ed616e1ccd0f39c0. Thanks for your contribution.

PR #140, opened 2019-05-06
Login twice issue

stanleybz:
#717223 - Users have to login twice.
#114165 - LS Cache making users login twice
#413338 - Wordpress login issue with Litespeed cache

The reason why login twice bug occurred after 2.9.5 is because of `LiteSpeed_Cache_Mediacan_media()`
will run `LiteSpeed_Cache_Config::get_instance()->in_exclude_optimization_roles()`
which run `LiteSpeed_Cache_Router::get_role()`,

In the get_role call the function cannot get correct uid might because
in_exclude_optimization_rolesis a static call because `LiteSpeed_Cache_Config::get_instance()`,

so when `LiteSpeed_Cache_Vary::finalize_default_vary()` run `get_role($uid)`,
the get_role already cached the role on LITESPEED_WP_ROLE variable and return it directly,
which make finalize_default_vary cannot set vary.

adding check `is_null( $uid )` on get_role can make the call on `finalize_default_vary()`
return a fresh value instead of cached value.

### DIFF
--- a/litespeed-cache/inc/router.class.php
+++ b/litespeed-cache/inc/router.class.php
@@ -214,7 +214,7 @@ class LiteSpeed_Cache_Router
 	 */
 	public static function get_role( $uid = null )
 	{
-		if ( defined( 'LITESPEED_WP_ROLE' ) ) {
+		if ( defined( 'LITESPEED_WP_ROLE' && is_null( $uid ) ) ) {
 			return LITESPEED_WP_ROLE ;
 		}
 


hailite:
We can't fix this way.

1) The 2nd condition  will cause double define `LITESPEED_WP_ROLE` which will be a fatal error. If you didn't meet it, please check ur dev env and fix it.

2) This issue happened bcos we check user role in `media.cls::constructor` in hook `after_setup_theme` which is before `init` hook. While user info is initialized in `init`  right after `after_setup_theme`. I will go through our init process and move all user related init to use the hook `init`.

stanleybz:
Noted, please let me know if need help, thanks.

hailite:
Fixed in https://github.com/litespeedtech/lscache_wp/commit/69e7bc71d0de5cd58961bae953380b581abdc088

PR #141, opened 2019-05-13
Quick fix for get_item() on multisite.

stanleybz:
#279467 - Lazy Loading Bug with Multisite 

Provide a quick fix 2 months ago, remember we discussed will do the fix on another way but seems not applied yet, provide a quick fix version first.


hailite:
This fix could result in other issues. It will cause all item settings use the primary setting, regardless that use primary conf setting. Its already fixed in v3.0 refactoring.

PR #148, opened 2019-06-07
Fix facetwp FWD.reset error on woocommerece page

stanleybz:
For Facetwp, if the template is "wp", return the buffered HTML with wp_send_json, so set ESI params `is_json = 1`


hailite:
Improved in https://github.com/litespeedtech/lscache_wp/commit/0b0cffbd8becd50346f9906c4ce05457dd4ee6e0 based on these 2 commits. Merged there.

hailite:
@stanleybz Please be aware of that this can NOT pass php5.3 validation and `&& ! empty( $_POST['data']['template'] === 'wp' )` is not a good way in programming style also.

PR #152, opened 2019-07-11
add null onload handler once it is used

joejordanbrown:
[loadCSS](https://github.com/filamentgroup/loadCSS) recommends nulling the onload handler once it is used since some browsers will occasionally re-call the handler upon switching the rel attribute to the stylesheet.

### DIFF
--- a/litespeed-cache/inc/optimize.class.php
+++ b/litespeed-cache/inc/optimize.class.php
@@ -375,7 +375,7 @@ class LiteSpeed_Cache_Optimize
 
 						$snippet = '' ;
 						foreach ( $urls as $url ) {
-							$snippet .= "<link rel='preload' data-asynced='1' data-optimized='2' as='style' onload='this.rel=\"stylesheet\"' href='$url' />" ;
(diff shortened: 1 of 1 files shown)


hailite:
Is there any reason that replace single quote to double quotes?

joejordanbrown:
Sorry, that's from our internal style guide rules. 

> PHP: 
Use `"` double quote for consistency and interpolation, this is the most common PHP style guide recommendation.

> HTML: 
Use `"` double quote  for consistency, this is the most common HTML style guide recommendation.

> JS: 
Use `'` single for consistency, this is the most common JS style guide recommendation.
> 

I can amend, do you have a style guide? 

I noticed there was a mix of " and ' quotes used for PHP. It might be wise for the project to include a [phpcs](https://github.com/squizlabs/PHP_CodeSniffer) ruleset config file that we can run to keep the code consistent. WordPress has standards are [here](https://github.com/WordPress/WordPress-Coding-Standards).

hailite:
It depends on which can make codes easier to read currently :)

PR #158, opened 2019-07-29
FIxing child images not update or destory postmeta bug.

stanleybz:
#167713 - LiteSpeed Cache Image Optimization stuck.

1) $root_id_list on 854 is an array, should merge with the original array.
2) optm_info on row 859 ($child_postmeta_info[ $v->root_id ] ) should serialize before storing to db otherwise will cause sql error.


hailite:
This is a good catch!

PR #172, opened 2019-09-04
ESI menu converting for Buddy Press

stanleybz:
[Asana] [3rd] BuddyPress compatible issue
https://wordpress.org/support/topic/support-of-members-connected-buddypress/

Converting menu as ESI with `private,no-vary` to show correct variable on menu.


hailite:
@stanleybz Didn't see client reply in forum topic. Also not receiving @ ruikai's test approval.

PR #175, opened 2019-10-03
Fixing amp bypass bug.

stanleybz:
Use `function_exists( 'is_amp_endpoint' )` to detect AMP installed.

Mimicking original `is_amp_endpoint()` to check `is_admin` and amp query string by `$_GET` instead of wp_query.

"/amp" will redirect to "?amp".

### DIFF
--- a/litespeed-cache/thirdparty/lscwp-3rd-amp-bypass.cls.php
+++ b/litespeed-cache/thirdparty/lscwp-3rd-amp-bypass.cls.php
@@ -26,7 +26,9 @@ class LiteSpeed_Cache_ThirdParty_AMP_Bypass
 	 */
 	public static function pre_load()
 	{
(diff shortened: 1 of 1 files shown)


litespeed-cache/thirdparty/lscwp-3rd-amp-bypass.cls.php
```diff
@@ -26,7 +26,9 @@ class LiteSpeed_Cache_ThirdParty_AMP_Bypass
 	 */
 	public static function pre_load()
 	{
-		if ( ! function_exists( 'is_amp_endpoint' ) || ! is_amp_endpoint() ) return ;
+		if ( ! function_exists( 'is_amp_endpoint' ) && is_admin() ) return ;
```
hailite: Are you sure this is correct?

PR #207, opened 2020-04-02
[GUI] Many UI updates & clarifications

J-Rey:
Please see each commit for detailed descriptions


tpl/page_optm/settings_media.tpl.php
```diff
@@ -40,7 +40,7 @@
 		<td>
 			<?php $this->build_input( $id, 'litespeed-input-long' ); ?>
 			<div class="litespeed-desc">
-				<?php echo __( 'Specify a base64 image to be used as a placeholder while other images finish loading.', 'litespeed-cache' ); ?>
+				<?php echo __( 'Specify a base64 GIF image to be used as a simple placeholder while images finish loading.', 'litespeed-cache' ); ?>
```
hailite, line 43: Doesn't have to be a GIF.

tpl/page_optm/settings_media.tpl.php
```diff
@@ -150,7 +150,7 @@
 		<td>
 			<?php $this->build_switch( $id ); ?>
 			<div class="litespeed-desc">
-				<?php echo sprintf( __( 'Automatically generate %s in the background via a cron-based queue.', 'litespeed-cache' ), __( 'Responsive Placeholder', 'litespeed-cache' ) ); ?>
+				<?php echo sprintf( __( 'Automatically generate %s in the background via a cron-based queue.', 'litespeed-cache' ), __( 'LQIP', 'litespeed-cache' ) ); ?>
```
hailite, line 153: As LQIP is a special word, prolly no translation, no need sprintf anymore

tpl/page_optm/settings_media.tpl.php
```diff
@@ -70,8 +70,8 @@
 		<td>
 			<?php $this->build_input( $id, 'litespeed-input-long' ); ?>
 			<div class="litespeed-desc">
-				<?php echo __( 'Specify an SVG to be used as a placeholder when generating locally.', 'litespeed-cache' ); ?>
-				<?php echo __( 'It will be converted to a base64 placeholder on-the-fly.', 'litespeed-cache' ); ?>
+				<?php echo __( 'Specify a SVG to be used as a placeholder when generating locally.', 'litespeed-cache' ); ?>
```
lslisa, line 73: I say "SVG" as the letters S V and G, so "an" would be appropriate here.

Is there a more standard way to say SVG that would make "a" a better choice?

tpl/page_optm/settings_media.tpl.php
J-Rey: Oh okay

tpl/page_optm/settings_media.tpl.php
J-Rey: Good point, yeah

tpl/page_optm/settings_media.tpl.php
J-Rey: No, I was thinking that since it started with a consonant but after looking it up then its more on how it sounds like & "es" is more how we pronounce it so you're right.

PR #214, opened 2020-04-27
Should verify peer and host in production

lucasRolff:


hailite:
We can't do this way. It changed by purpose long time ago bcos some users have invalid or error in their cert.

lucasRolff:
For crawler, maybe, if the user has cert issues, then the issues should be resolved by the user, or allow the user to be able to set a checkbox to ignore all security.

For `Cloud::` it's not the users problem. You should **not** communicate unverified to any service.

Defaults for SSL communication should **never** be "ignore certificate errors".

PR #222, opened 2020-05-11
Update PHP requirement to 5.6.

JTS-FIN:
Cache > 3.0 requires PHP 5.6, due to this file atleast:
https://github.com/litespeedtech/lscache_wp/blob/master/src/task.cls.php

You can see from here, that it fails if PHP version is less than 5.6:
https://3v4l.org/5eQXt#v560

So we update the requirement, to keep it from trying to run and to
fail in PHP version < 5.6.0.


hailite:
Thanks for the report. We need to maintain PHP5.3's compatibility. Will fix this in next release.

PR #223, opened 2020-05-14
Fix Script URI Issue in 3.0.9

aonsyed:


hailite:
Fixed in https://github.com/litespeedtech/lscache_wp/commit/17c2022196e9162a9b36bb00ed295dadc13ba57a Thanks.

Issue #227, opened 2020-05-22
data/*.ini files are public when using OpenLiteSpeed Web Server


spenweb:
Apache directives are not recognized in .htaccess files by OpenLiteSpeed Web Server (as described here: https://openlitespeed.org/kb/access-control/). To have these `litespeed-cache/data/*.ini` files non-accessible, an Apache rewrite should be used as well in the `litespeed-cache/data/.htaccess` file. I will make a PR to correct this bug.

spenweb:
Here is the proposed PR to fix this bug: https://github.com/litespeedtech/lscache_wp/pull/228

PR #228, opened 2020-05-22
Deny access with Apache rewrite rule

spenweb:
Supports OpenLiteSpeed Web Servers. OLSWS does not support Apache directives in .htaccess files. Apache rewrite rules are supported and thus should be used to properly deny access to the `data/` directory. See this support article for more information: https://openlitespeed.org/kb/access-control/.

### DIFF
--- a/data/.htaccess
+++ b/data/.htaccess
@@ -1,3 +1,7 @@
 Order Deny,Allow
-
 Deny from All
+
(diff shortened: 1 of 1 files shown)


spenweb:
Here is the related issue: https://github.com/litespeedtech/lscache_wp/issues/227

PR #246, opened 2020-06-26
Various ESI nonce updates & minor readme revisions

J-Rey:


hailite:
The ESI nonce list will be uncommented later. Need to drop the 3rd party code in the same time

PR #253, opened 2020-07-19
Spelling and grammar updates

blastoise186:
Hi,

I've been using LiteSpeed for a couple of years now and I've often noticed a few spelling and grammar mistakes in the WordPress plugin. I've had a look through and tried to catch as many as I could possibly find in one go. While I did rewrite a few things, I've tried to keep them as close as possible to the original versions

There's a few things that I noticed which sounded like I could have improved them, but I wasn't sure about the best way to do so. I've left those ones alone for the time being. I'm also unsure about whether to update the file headers or not, so I left those untouched.

Thanks!


blastoise186:
Turns out I forgot to update the strings in a whole bunch of other places. Let me fix those! I'm going to mark this as a Draft for a bit while I catch up.

blastoise186:
Ok! I _think_ I've caught up with the entire set. Sorry for the flood of commits!

Now that I've gone through the entire set in the original locations, I realise it might not have been necessary to edit the first file I modified in this PR. However, I decided to leave it as-is for now in case it helps.

lslisa:
@hailite I've reviewed this PR.

I would prefer *not* to change `LiteSpeed Enterprise Web Server` to `LiteSpeed Web Server Enterprise` on line 2186.

Also, I agree that line 2674 is badly worded, but I don't like the proposed change. Let's do this:

```
"When enabled, the cache will automatically purge when any plugin, theme "
"or the WordPress core is upgraded."
```

Otherwise, I am fine with the rest of the suggestions.

blastoise186:
Sure thing, bear with me for a few minutes and I'll make those tweaks.

blastoise186:
Thanks for the feedback! I've gone ahead and updated those items as suggested.

lslisa:
Looks good to me! Thanks for the PR @blastoise186 :-)
Over to you, @hailite

blastoise186:
You're welcome :)

I've used LiteSpeed for a while now and I've always felt it was a bit strange to have all those spelling errors. I'm glad to be able to help fix those. If I ever spot more, I'll let you know.

blastoise186:
Ok. It looks like I need to catch up. Bear with me on this!

blastoise186:
Ok. It looks like some of these conflicts were caused by factors outside my control and I'm not sure if I can fix them directly myself.

I've gone ahead and enabled the ability for maintainers to edit this PR, since I have a feeling that I'm likely to make a mistake if I try to fix the conflicts myself

hailite:
Thanks for your PR. Merged. Apologize for the delay.

If you have new PR, you can just modify the files, no need to update the .pot file as it is automatically generated in each release.

blastoise186:
Thank you!

I'll be sure to drop by again if I come across any others. And thanks for letting me know about that .pot file. At very least, it makes it easy for me to find these things. :)

PR #267, opened 2020-09-18
enable webp replacement on iOS 14

qtwrk:


hailite:
https://wordpress.org/support/topic/webp-ios14-support/#post-13423788

PR #274, opened 2020-11-09
Adding Post Max Age-logics for Sitemap-gen

exetico:
I've request a Post Max Age-logic for the Sitemap gen, more than a few mounts ago. After checking the current setup I realised how simple the plugin-structure is build up. That's great news!

Here's my take on a Post Max Age-logic.

I've already testet it on the latest version of WP with LiteSpeed server and LS Cache.

Here's one example with 90 days logics:
image

Here's another with 0 (= disabled):
image

Here's the "post"-overview for reference. Please note that the dates is listed as DD-MM-YYYY on the screenshot:
image

### DIFF
--- a/lang/litespeed-cache.pot
+++ b/lang/litespeed-cache.pot
@@ -3237,6 +3237,14 @@ msgstr ""
 msgid "These options will be invalid when using %s."
 msgstr ""
 
+#: tpl/crawler/settings-sitemap.tpl.php:120
+msgid "Posts Max Age"
+msgstr ""
+
+#: tpl/crawler/settings-sitemap.tpl.php:126
+msgid "Posts older than max age (in days), will not be included in the sitemap."
+msgstr ""
+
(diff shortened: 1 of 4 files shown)


hi-hai:
Thanks for your PR. I believe most of current users are using 3rd party plugin to generate the sitemap, I am not satisfied w/ the current LSCWP internal sitemap generation. Will see in future if can have a chance to improve it or not. Will come back to this PR at that moment. Thanks.

I will note this as closed for now.

exetico:
Hi Hai

If you're not happy with the current solution, why don't you allow the current request to pass the gate? :-) 

I can't find any good reason for not allowing it. 

Else, I'd need to create a custom plugin, just to implement another solution like this. That doesn't make sense, as seen from my POV.

A new and more refined solution would be great to have, but I can't see why it's the dealbeaker? Not at all :-)

PR #295, opened 2021-01-28
Dev

usabe:


hi-hai:
CDN parse image src issue can't be fixed like this. The correct direction is to fix the regex to parse the correct image src.

You can pass this to client as a tmp fix. This can't be in public versions.

PR #329, opened 2021-04-13
Preserve the vary value before appending 'ismobile'

usabe:
@devmanishtomar @qtwrk


hi-hai:
Duplicated of https://github.com/litespeedtech/lscache_wp/pull/328

PR #335, opened 2021-04-14
Update JS Minify lib

usabe:


usabe:
@luminsol : https://wordpress.org/support/topic/js-minification-removes-spaces-from-template-literals/

PR #338, opened 2021-04-16
auto-change port to 11211/6379

qtwrk:


hi-hai:
If user customized the port, then toggle the option, better to not auto change the port. That was the reason I dropped the dynamic change way.

```
Jonathan Kagan
Also maybe grey out memcached or redis if it cannot detect the extension or the package on the system.
```
Jon's idea is cool. Will do this way.

Thanks for the commit though.

PR #343, opened 2021-05-14
Fix error with notify

lucasRolff:
Change missing `$_POST` to `$post_data` so DB status gets updated as it should


hi-hai:
Please always PR to dev.

PR #346, opened 2021-05-15
Improves AMP compatibility.

milindmore22:
PR Addresses following issue

- Fixes Additional NoScript tag being injected in amp-img > img tag more details in [this](https://wordpress.org/support/topic/noscript-enclosure-for-boilerplate-is-noscript/) support forum.
- Disables Page Optimizations on AMP pages (Combining and Minifying HTML, JS, CSS)
- Disables Lazy loading image and iframes as AMP already does that.
- Improves AMP compatibility so that LiteSpeed Cache plugin works flawlessly on all [template modes](https://amp-wp.org/documentation/getting-started/template-modes/) (Standard Mode, Transitional Mode, and Reader Mode)


hi-hai:
The above issues should have been fixed in latest dev version. Can you give it a test? Thanks.

milindmore22:
Hello @hi-hai 

I tested the latest dev but seems the issue is not resolved, here is a [demo url](https://plugins.dev6.rt.gw/amp/) I turned all Page optimizations ON but they are not bypassed on the AMP endpoint as you can see they are marked by the [AMP validator ](https://validator.ampproject.org/#url=https%3A%2F%2Fplugins.dev6.rt.gw%2Famp%2F)

I think the issue here is AMP detection logic which needs to be improved. you can use the [amp_is_request()](https://amp-wp.org/reference/function/amp_is_request/) function to detect AMP endpoint the function is available post `plugins_loaded`, but it safe to call them on `wp` as per my PR

let me know If my PR needs any further improvements

hi-hai:
Couple things to improve:
1. Commit to `dev` branch.
2. Follow the original logic to run the detection in `preload` directly. That `preload` is fired in hook `after_setup_theme`.
3. Same as point 2). Don't change the const definitions. The short way is preferred. Also your PR missed Guest Mode definition.

milindmore22:
> Commit to dev branch.

OK, I will recreate a new branch from `dev`

> Follow the original logic to run the detection in preload directly. That preload is fired in hook `after_setup_theme`.

The `amp_is_request()` function can provide better AMP detection and it is available post `wp` hook so the AMP detection logic needs improvement. [reference](https://codex.wordpress.org/Plugin_API/Action_Reference) 

The following Template Modes are impacted: 

- Standard Mode (AMP fist Site): https://example.com/
- Transitional Mode:  path suffix: https://example.com/page/amp/ 
- Reader Mode: Legacy Reader Mode posts

image


> Same as point 2). Don't change the const definitions. The short way is preferred. Also your PR missed Guest Mode definition.

OK no issue I followed the [WordPress Coding Standards](https://github.com/WordPress/WordPress-Coding-Standards)

hi-hai:
Thanks for your detailed explanation.

Our all optimizations are initialized in hook `init`. This means if we fired amp detection in `wp`, some optimizations will still be kicked off. However, as the major issue is about the Guest Mode script. That script is inserted in optimization finalization, which is in `litespeed_buffer_finalize` hook. This hook is in the very end, when the buffer is sending out. So `wp` hook should work.

hi-hai:
If the amp detection can't be done in preload, let's use `wp` then.

PR #351, opened 2021-05-22
PR: Fix for 'No valid sitemap parsed for crawler' errors

appnostic-io:
This is a fix for errors stating 'No valid sitemap parsed for crawler' errors.

This has been tested on OLS 1.7.10 and latest master of the cache plugin.

The error seems to originate because the PHP SimpleXMLElement object is not explicitly cast.  Thus to cast it to string results in the proper operation of the '! empty' checks as well as the correct population of $this->_urls used in other parts of the class.

Please let me know if there is anything else I need to do.

Thank you for a most excellent plugin.  You're a credit to open source.


hi-hai:
The fix is good. Can you recreate the PR to `dev` branch? PRs can only be accepted in `dev`. Thanks.

PR #352, opened 2021-05-29
Image optimization

lowwebtech:
Hi,

I just did some image optimization to reduce the file size of the images, mostly images of screenshots.
File size of screenshots are reduced by 2.96MB !!!

PNG images are converted to PNG8 (256 colors often less) and optimized with ImageOptim.

Also, these screenshots are bundled inside plugin. 
File size of the zip : 4.9MB
zip without screenshots : 840kB

You can add a `.distignore` file with wp_assets folder in it, look at [wp dist-archive](https://developer.wordpress.org/cli/commands/dist-archive/)

Hope it helps


hi-hai:
Well done! Thanks.

hi-hai:
`. distignore` is a good idea. Or maybe can use `svn:ignore`. The very beginning purpose to include wp_assets in git is to make the management easier. There is indeed no reason to include them in distribution. 

Will exclude them from future releases since v4.1, both WP packages and Beta Test. Thanks.

PR #362, opened 2021-06-23
UI/various revisions; added more CDN mapping default file types & nonces

J-Rey:


J-Rey:
Sorry its messy. Cleaning up the history made more commits but at least I got it so the diffs are reasonable now.

hi-hai:
Good job! Thanks.

PR #365, opened 2021-06-26
Add wsform nonce

cstrouse:
Add default nonce for wsform (https://wsform.com/)


westguard:
I believe this should be:

wsf_post

wsf_nonce is the input name we use to identify the nonce. wsf_post is the action passed to the wp_create_nonce function.

Thanks!

Mark

PR #371, opened 2021-07-10
compatibility with Gravity Forms 

qtwrk:
compatibility with Gravity Forms

### DIFF
--- a/thirdparty/entry.inc.php
+++ b/thirdparty/entry.inc.php
@@ -24,7 +24,6 @@ $third_cls = array(
 	'Divi_Theme_Builder',
 	'Elementor',
 	'Facetwp',
-	'Gravity_Forms',
 	'Theme_My_Login',
 	'User_Switching',
 	'WCML',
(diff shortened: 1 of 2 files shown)


thirdparty/entry.inc.php
```diff
@@ -48,3 +49,4 @@
 add_action( 'litespeed_init', 'LiteSpeed\Thirdparty\NextGenGallery::preload' );
 add_action( 'litespeed_init', 'LiteSpeed\Thirdparty\AMP::preload' );
 add_action( 'litespeed_init', 'LiteSpeed\Thirdparty\Elementor::preload' );
+add_action( 'litespeed_init', 'LiteSpeed\Thirdparty\Gravity_Forms::preload' );
```
hi-hai, line 52: You can't define both `detect` and `preload` but only have `preload` defined in the following class.

thirdparty/gravity-forms.cls.php
```diff
@@ -0,0 +1,26 @@
...
+defined( 'WPINC' ) || exit;
+
+class Gravity_Forms
+{
+    /**
+     * Check if GF is enabled and disable LSCWP on gf-download and gf-signature URI
+     *
+     * @since 4.1.0 #900899 #827184
+     */
+    public static function preload()
```
hi-hai, line 18: Try to rename this function to `detect` and drop the `add_action` above to see if it can work or not.

PR #379, opened 2021-07-25
webp replacement for blogs.dir

qtwrk:


hi-hai:
We won't consider wp 3.5 support. So this will be no more used

PR #380, opened 2021-07-27
LQIP Network Compatibility | Purge Only UCSS | [Root] clear_q | Debug Log Append [Class]

alicetang2cool-github:


src/avatar.cls.php
```diff
@@ -270,7 +240,7 @@ private function _generate( $url ) {
 		self::save_summary();
 
 		// Generate
-		if ( ! self::has_cache() ) {
+		if ( ! $this->has_cache_folder( 'avatar', $this->_is_subsite_purge() ? get_current_blog_id() : false ) ) {
```
hi-hai, line 243: this needs to use purge-> _is_subsite_purge()

hi-hai: question: why create the folder?

PR #386, opened 2021-08-09
LQIP Network Compatibility & clear_q

alicetang2cool-github:


src/avatar.cls.php
```diff
@@ -270,7 +240,7 @@ private function _generate( $url ) {
 		self::save_summary();
 
 		// Generate
-		if ( ! self::has_cache() ) {
+		if ( ! $this->has_cache_folder( 'avatar', $this->cls('Purge')->_is_subsite_purge() ? get_current_blog_id() : false ) ) {
```
hi-hai, line 243: This is to generate avatar, shouldn't use `_is_subsite_purge` function to detect. As you are using the blog_id detection in https://github.com/litespeedtech/lscache_wp/pull/386/files#diff-9456edaa82557d2a23d04e2abe2b665a0a8c31255da28f97fbe5d30ca21a07daR383 too, maybe you can extract them to a function.

src/gui.cls.php
```diff
@@ -499,7 +501,7 @@ public function frontend_shortcut() {
 			) );
 		}
 
-		if ( $this->cls( 'CSS' )->has_ccss_folder() ) {
+		if ( $this->has_cache_folder( 'ccss', $subsite_id ) ) {
```
hi-hai, line 504: For what reason did you pass in `$subsite_id` while not directly generate it in the function definition? Won't it be easier as you don't need to pass each time?

alicetang2cool-github: Ah I see, ok I will pass it in the has_cache_folder

src/avatar.cls.php
alicetang2cool-github: got it!

alicetang2cool-github: I did this 
`// Generate
		if ( ! $this->has_cache_folder( 'avatar' ) ) {
			$this->_mkdir();
		}`
and for Root class
`public function has_cache_folder( $type ) {
		$subsite_id = is_multisite() && ! is_network_admin() ? get_current_blog_id() : '';

		if ( file_exists( LITESPEED_STATIC_DIR . '/' . $type . '/' . $subsite_id ) ) {
			return true;
		}
		return false;
	}`

PR #387, opened 2021-08-09
LQIP Network Compatibility & clear_q

alicetang2cool-github:


src/placeholder.cls.php
```diff
@@ -250,7 +250,7 @@ private function _placeholder( $src, $size ) {
 		$tmp_placeholder = $this->_generate_placeholder_locally( $size );
 
 		// Store it to prepare for cron
-		$queue = $this->_load_queue( 'lqip' );
```
hi-hai, line 253: The reason this didn't use `$this->_queue` like CCSS/UCSS is just it is historic code and didn't get the chance to refactor yet. After I approved this PR, if you are interested, you can consider to change it and drop the 2nd param of `save_queue()` func.

src/root.cls.php
```diff
@@ -19,7 +19,52 @@ abstract class Root {
...
+		File::rrmdir( LITESPEED_STATIC_DIR . '/' . $type . '/' . $subsite_id );
+
+		// Clear All summary data
+		$this->_summary = array();
+		self::save_summary();
+
+		if ( $type == 'ccss' || $type == 'ucss') {
+			Debug2::debug( '[CSS] Cleared ' . $type .  ' queue' );
+		}
+		elseif ( $type == 'avatar' ) {
```
hi-hai, line 56: Isn't the debug log in this condition same as the `else{}` one?

hi-hai: This is not a big deal but can be improved still. I will approve this PR. You can include this improvement in your next PR.

PR #391, opened 2021-08-11
Fix v4 upgrade cleanup, Increase Object Caching & UI updates

J-Rey:
More global object cache groups
Fix affecting v4 upgrade cleanup
Small UI updates

### DIFF
--- a/src/data.upgrade.func.php
+++ b/src/data.upgrade.func.php
@@ -38,8 +38,8 @@ function litespeed_update_4() {
 	$q = 'DROP TABLE IF EXISTS ' . $tb;
 	$wpdb->query( $q );
 
-	if ( file_exists( LITESPEED_STATIC_DIR . '/ccss' ) ) {
-		File::rrmdir( LITESPEED_STATIC_DIR . '/ccss' );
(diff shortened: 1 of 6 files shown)


src/data.upgrade.func.php
```diff
@@ -38,8 +38,8 @@ function litespeed_update_4() {
 	$q = 'DROP TABLE IF EXISTS ' . $tb;
 	$wpdb->query( $q );
 
-	if ( file_exists( LITESPEED_STATIC_DIR . '/ccss' ) ) {
-		File::rrmdir( LITESPEED_STATIC_DIR . '/ccss' );
+	if ( file_exists( LITESPEED_STATIC_DIR . '/ccsjs' ) ) {
```
hi-hai, line 41: This is a good catch. However it won't work for v4 existing users. Can put this into v4.3 upgrade func.

hi-hai: Done in https://github.com/litespeedtech/lscache_wp/commit/9d7baa0bc799792c22b352c976eb83184c74f8c1

PR #402, opened 2021-09-16
bypass js optm if data-cfasync="false"

qtwrk:


galbaras:
Any idea when this will be merged? Once it is, I can start using the plugin and you can boast plugin and Cloudflare compatibility.

galbaras:
This really seems like an easy win. Any reason for not releasing it?

hi-hai:
@galbaras Do you have any idea which plugin added that attribute?

galbaras:
Several popular plugins add this attribute to their scripts, like MonsterInsights and EWWW Image Optimizer, but it can also be added in themes or via filters, etc.

Why would you not do it? This attribute means that the script developer wanted it skipped by the Cloudflare page optimiser, which implies ANY page optimiser.

When I suggested it to the developer of WP Fastest Cache, he added it, I tested it, and he released it, so now sites using that plugin for optimisation run more correctly, even if they load slightly more slowly.

hi-hai:
This makes sense. Thanks for explanation.

galbaras:
When will this be released?

galbaras:
Has this been released yet?

PR #419, opened 2021-11-20
Update esi.nonces.txt

qtwrk:


hi-hai:
As there is conflict, I will commit w/ a new one

PR #430, opened 2022-01-29
Force gravatar cache refresh in browsers and on Cloudflare

### LINKED ISSUE #322: Gravatar Cache vs. browser cache and Cloudflare
When someone changes the avatar on my site, after clearing the Gravatar Cache, the browsers still display the old version. But even clearing the browser cache does not help, because there is an old version of the avatar sitting in the Cloudflare cache.

After each clearing of Gravatar Cache, a random query string should be added to URL just like it is with CSS and JS.

rafaucau:
Resolves #322 

This PR adds a query string `?ver` with the date the gravatar was saved. This way, when a user changes their avatar, it will be immediately visible after clearing the gravatar cache. No more need to manually purge the cache on Cloudflare and in the browser.


hi-hai:
Thanks

PR #437, opened 2022-02-15
Rename blacklist to disallowed list in GUI.

### LINKED ISSUE #436: Rename blacklist to disallowed list in GUI.
WordPress core went with `disallowed list` instead of `blacklist` in 5.5.
https://make.wordpress.org/core/2020/07/23/codebase-language-improvements-in-5-5/

Maybe it's to much work to change the whole code base, but at least user facing strings could be updated in my opinon (and then also docs.)

kebbet:
Closes #436 436


hi-hai:
Thanks.

PR #438, opened 2022-02-15
Rename whitelist in GUI

kebbet:
Follow up to #437


hi-hai:
Per the discussion w/ Lisa, we will use `allowlist` for consistency.

PR #451, opened 2022-04-04
Add ESI shortcode support for silence parameter

joyfullservice:
This change allows you to add the `_ls_silence=true` parameter to the `esi` shortcode modifier to prevent the output of ESI comments in the shortcode output. This is especially helpful when using shortcodes in javascript.

As a case in point, we are using ESI with a shortcode that generates a custom URL for a media item played through JWPlayer. Since the shortcode is intended to return a plain URL, the normal ESI comments then cause the URL to become invalid.

```php
<script type="text/javascript">
  jwplayer("contentvideo_block").setup({
    id: "jwplayervideohtml",
    file: "[esi hls_url filename='[types field='video-filename'][/types]'",
```

While the output contains a valid HTML comment, this comment breaks the javascript property assignment:

image

With this proposed change, I can simply adjust my shortcode to add the new parameter:

image

The output then does not include the comments, and everything works just as I would expect.

I believe this would benefit other users as well in cases where you need to skip the comment output in specific shortcodes without having to disable them at the global level.

Thanks for your consideration!!  😄

### DIFF
--- a/src/esi.cls.php
+++ b/src/esi.cls.php
@@ -210,7 +210,7 @@ class ESI extends Root {
 		}
 
 		$silence = false;
-		if ( ! empty( $atts[ '_ls_silence' ] ) && ( $atts[ '_ls_silence' ] ) ) {
+		if ( ! empty( $atts[ '_ls_silence' ] ) ) {
 			$silence = true;
 		}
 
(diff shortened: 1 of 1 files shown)


robertstaddon:
This would be quite helpful functionality to add! While the LiteSpeed WordPress plugin allows the outputting of comments to be globally disabled through the "LSCACHE_ESI_SILENCE" constant in wp-config.php, there are many situations where it would be helpful to silence only a specific esi shortcode (outputting into JavaScript or JSON, for example). This "silence" parameter has been conspicuously missing from the esi shortcode up to this point and this simple patch does a nice job of fixing it!

hi-hai:
Thanks for your contribution. This will be in next release v4.7

PR #468, opened 2022-06-13
Add an incompatible plugin notification

tynanbe:
Incompatible plugin notification:
- Can be dismissed
- Is refreshed when any plugin is (de)activated
- Is refreshed when `litespeed-cache` itself is activated

### DIFF
--- a/src/activation.cls.php
+++ b/src/activation.cls.php
@@ -69,6 +69,11 @@ class Activation extends Base {
 		if ( defined( 'LSCWP_REF' ) && LSCWP_REF == 'whm' ) {
 			GUI::update_option( GUI::WHM_MSG, GUI::WHM_MSG_VAL );
 		}
+
+		Thirdparty\LiteSpeed_Check::activated_plugin(
+			basename( LSCWP_DIR ) . '/litespeed-cache.php',
+			null
+		);
 	}
 
 	/**
@@ -253,6 +258,11 @@ class Activation extends Base {
 
 		// delete in case it's not deleted prior to deactivation.
 		GUI::dismiss_whm();
+
+		Thirdparty\LiteSpeed_Check::deactivated_plugin(
+			basename( LSCWP_DIR ) . '/litespeed-cache.php',
+			null
+		);
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


src/activation.cls.php
```diff
@@ -69,6 +69,11 @@ public static function register_activation() {
 		if ( defined( 'LSCWP_REF' ) && LSCWP_REF == 'whm' ) {
 			GUI::update_option( GUI::WHM_MSG, GUI::WHM_MSG_VAL );
 		}
+
+		Thirdparty\LiteSpeed_Check::activated_plugin(
```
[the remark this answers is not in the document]
tynanbe: Will refactor to use detect/set/clear an option instead.

thirdparty/litespeed-check.cls.php
```diff
@@ -0,0 +1,151 @@
...
+
+		/**
+		 * Check for incompatible plugins when any plugin is (de)activated.
+		 */
+		add_action( 'activated_plugin', __CLASS__ . '::activated_plugin', 10, 2 );
+		add_action( 'deactivated_plugin', __CLASS__ . '::deactivated_plugin', 10, 2 );
+	}
+
+	public static function activated_plugin( $plugin, $network_wide ) {
+		self::incompatible_plugin_notice( $plugin, $network_wide, 'activated' );
```
hi-hai, line 69: If the plugin is not in the incompatible list, may not need to call `incompatible_plugin_notice()`. Bcoz each call will cause 2 db operations right?

hi-hai: It may also be good to show the msg each time when any plugin is installed

tynanbe: > If the plugin is not in the incompatible list, may not need to call `incompatible_plugin_notice()`. Bcoz each call will cause 2 db operations right?

For future reference, it should be simple to add an early exit condition that detects whether the plugin that was just activated is in the incompatible plugins list. This check would be the first thing in the `incompatible_plugin_notice` function.

tynanbe:
Thank you!

PR #471, opened 2022-07-05
Don't correct hash-tag URLs when minifying css

### LINKED ISSUE #463: CSS Minify bug
✌ Hello there,

I found a bug that appears when “CSS Minify” is enabled,

1️⃣ Original CSS file:
`.some-selector {
	filter: url(#hash);
}`

2️⃣ CSS file after “CSS Minify”:
`.some-selector {
   filter: url(/wp-content/themes/theme-name/#hash)
}`

tynanbe:
As discussed in: https://wordpress.org/support/topic/css-minify-combine-breaks-clip-path/

CSS rules like this:

```css
.some { property: url(#my-object); }
```

were rewritten and minified like this:

```css
.some{property:url(/wp-content/themes/twentytwentyone/#my-object)}
```

With this PR they're not rewritten anymore, and end up like this after minification:

```css
.some{property:url(#my-object)}
```

Closes #463.

### DIFF
--- a/lib/css-min/urirewriter.cls.php
+++ b/lib/css-min/urirewriter.cls.php
@@ -291,8 +291,8 @@ class UriRewriter
             return $m[0];
         }
 
-        // if not root/scheme relative and not starts with scheme
-        if (!preg_match('~^(/|[a-z]+\:)~', $uri)) {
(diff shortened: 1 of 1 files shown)


tynanbe:
Thank you~

PR #473, opened 2022-07-26
Update ucss_whitelist.txt

Piirka:
Fixes Avada mobile menu. Avada creates mobile menu on-the-fly via Javascript. That's why class names are not showing up in source code.

### DIFF
--- a/data/ucss_whitelist.txt
+++ b/data/ucss_whitelist.txt
@@ -8,3 +8,7 @@
 
 ############# DIVI ################
 .et_pb_number_counter.active
+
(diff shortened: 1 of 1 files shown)


hi-hai:
Thanks for your PR. 

Question: If Avada generates those CSS via JS, how can that help UCSS as UCSS is gathered by PHP w/ the page HTML source?

Piirka:
You're welcome. 

Sorry, I meant in my case Avada generates different HTML output for mobile menu via JS.

PR #474, opened 2022-07-27
Fix to avoid fatal error on array merge

brunocantuaria:


brunocantuaria:
Ref: https://wordpress.org/support/topic/uncaught-typeerror-array_merge/#post-15862112

Snowbaha:
another fix is to check the array on line 583 :
```
$data_array = is_array($data) ? $data : [];
$existing_summary_array= is_array($existing_summary) ? $existing_summary : [];
$new_summary = array_merge( $existing_summary_array, $data_array );
```

hi-hai:
Thanks for your PR. Please commit to dev branch

We were thinking about that fix, but got the conclusion that it is not needed.

image

However, it is good to have tho. Will include it.

PR #482, opened 2022-09-15
Presets Feature

tynanbe:
Add a feature to choose from five standard configuration presets.

Backs up the current settings, with the ability to restore at any later time.


tynanbe:
Thank you!

PR #483, opened 2022-09-16
Fix: Let VPI work on the posts page

tynanbe:
This fix allows VPI functionality to work on a page set as the WordPress "Posts page," for which `is_singular()` returns `false`.


tynanbe:
Thank you!

PR #485, opened 2022-09-27
async lazyload lib

mam4dali:
Hi
lazy load In the latest version, it is no longer defer. this is good!
But it affects the total blocking time
So it is better to be async

### DIFF
--- a/src/media.cls.php
+++ b/src/media.cls.php
@@ -521,7 +521,7 @@ class Media extends Root {
 
 		// Include lazyload lib js and init lazyload
 		if ( $cfg_lazy || $cfg_iframe_lazy ) {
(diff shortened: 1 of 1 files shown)


hi-hai:
https://www.w3schools.com/tags/att_script_async.asp
> Note: The async attribute is only for external scripts (and should only be used if the src attribute is present).

I think `async` should be presented due to same reason why we dropped `defer`?

mam4dali:
> https://www.w3schools.com/tags/att_script_async.asp
> 
> > Note: The async attribute is only for external scripts (and should only be used if the src attribute is present).
> 
> I think `async` should be presented due to same reason why we dropped `defer`?

Apparently, I did not have the correct information about it
But I think it will work if we base64 it
Like other inline codes that are deferred with this method

mam4dali:
Unfortunately, lazy load still uses the onload event
In times when the internet is slow or a js is loaded with a delay, it causes the page loading to be greatly affected and it takes a long time to show the photos.

PR #489, opened 2022-10-21
Merge webp and webp_replace img_optm settings

tynanbe:


tpl/img_optm/settings.tpl.php
```diff
@@ -107,17 +108,13 @@
...
+				<br /><?php Doc::notice_htaccess(); ?>
+				<br /><?php Doc::crawler_affected(); ?>
 			</div>
 		</td>
 	</tr>
-
-	<?php
-		if ( ! is_multisite() ) :
-			// webp
-			require LSCWP_DIR . 'tpl/img_optm/settings.media_webp.tpl.php';
```
hi-hai, line 117: The reason we use separate file is for network settings as well. If you remove this file, in `img_optm/network_settings.tpl.php`, there will be no this option

tynanbe: Thanks, missed that. Should be fixed now.

tpl/general/settings.tpl.php
```diff
@@ -195,12 +195,6 @@
 				⚠️ <?php echo __( 'Notice', 'litespeed-cache' ); ?>: <?php echo sprintf( __( 'You need to turn %s on and finish all WebP generation to get maximum result.', 'litespeed-cache' ),  '<code>' . Lang::title( Base::O_IMG_OPTM_WEBP ) . '</code>' ); ?>
 				</font>
 				<?php endif; ?>
-
-				<?php if ( ! $this->conf( Base::O_IMG_OPTM_WEBP_REPLACE ) ) : ?>
-				<br /><font class="litespeed-danger litespeed-left10">
-				⚠️ <?php echo __( 'Notice', 'litespeed-cache' ); ?>: <?php echo sprintf( __( 'You need to turn %s on to get maximum result.', 'litespeed-cache' ),  '<code>' . Lang::title( Base::O_IMG_OPTM_WEBP_REPLACE ) . '</code>' ); ?>
-				</font>
-				<?php endif; ?>
```
hi-hai, line 203: I think we still need to give this notice

tynanbe: The notice is still given immediately above.

PR #495, opened 2022-11-03
[FEATURE] - Presets CLI

xLukii:
The following actions are allowed from the CLI:
- Obtain the list of generated backups.
- Restore a backup already created through the timestamp.
- Apply a Preset (Essentials, Basic, Advanced, Aggressive, Extreme)

This is a base, more options could be added in the future.


hi-hai:
Good job! Thanks.

PR #505, opened 2022-11-14
Delay purge when future posts get published

tynanbe:


src/purge.cls.php
```diff
@@ -488,7 +492,7 @@ private function _add( $tags, $purge2 = false ) {
 		}
 		else {
 			@header( $curr_built );
-			if ( defined( 'LITESPEED_DID_send_headers' ) && apply_filters( 'litespeed_delay_purge', false ) ) {
+			if ( defined( 'LITESPEED_DID_send_headers' ) || apply_filters( 'litespeed_delay_purge', false ) ) {
```
hi-hai, line 495: This change will cause all the normal visits sending purge2.

PR #515, opened 2022-12-17
Database optimizer now handles postmeta when cleaning revisions

tynanbe:
Fixes https://wordpress.org/support/topic/database-post-revisions-orphan-records/


IMTanuki:
Can you pls advise which WP plugin version handles this? 
5.3.1 is the latest ver for WP, released 2022-12-12
- AFAIK, it does not have this fix

Can you pls advise when a new plugin version will be available with this fix?

tynanbe:
It will likely be in the next release.
Until then, you might want to visit your `/wp-admin/admin.php?page=litespeed-toolbox#beta_test` page and choose to `Use latest GitHub Dev commit`.

PR #516, opened 2022-12-19
Fix object cache lib incr, desr functions (thanks bdrbros)

tynanbe:
Fixes https://wordpress.org/support/topic/internal-server-error-uncaught-error-call-to-undefined-function-incr_desr/


cbratschi:
Our system was running into this issue too. We just applied this patch until a new release is available. Thanks for this fix.

PR #534, opened 2023-02-14
Prevent undefined array key warning when widgets are disabled

tynanbe:
Fixes a warning when Widget functionality is completely removed.
```
PHP Warning: Undefined array key "WP_Widget_Recent_Posts" in .../litespeed-cache/src/purge.cls.php on line 1149
```

### DIFF
--- a/src/purge.cls.php
+++ b/src/purge.cls.php
@@ -1144,8 +1144,11 @@ class Purge extends Base {
 
 		global $wp_widget_factory;
 		// recent_posts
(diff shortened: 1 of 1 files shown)


tdwesten:
I have the same problem;
```
Warning: Undefined array key "WP_Widget_Recent_Posts" in /Users/thomasvanderwesten/server/moooicarpets/wp-content/plugins/litespeed-cache/src/purge.cls.php on line 1147
```

hi-hai:
@tdwesten V5.3.3 will include this fix. It will be packed today. You can use `Beta Test` to try latest dev for it.

PR #535, opened 2023-02-15
exclude jetpack stats JS

qtwrk:


hi-hai:
Please provide the reason to exclude

qtwrk:
reference: #322938

PR #545, opened 2023-04-12
Fix warnings for WP < v5.3

tynanbe:
Fixes warnings like:

```
[STDERR] PHP Warning:  include(.../wp-content/advanced-cache.php): failed to open stream: No such file or directory in .../wp-settings.php on line 84
```

A check for this file happens before plugins are loaded, and can omit this warning for any cache plugin.

Beginning with WordPress v5.3, this fix is not needed.


hi-hai:
Will merge after v5.4 rleased

PR #550, opened 2023-04-18
Ensure Object_Cache::delete returns a bool

### LINKED ISSUE #548: LS Object cache is not compatible with wp_cache_delete definition.
The [wp_cache_delete](https://developer.wordpress.org/reference/functions/wp_cache_delete/) function has to return a boolean.

`wp_cache_delete( int|string $key, string $group = '' ): bool`

LS Cache's object cache returns an integer. 

Relevant method:

https://github.com/litespeedtech/lscache_wp/blob/b41455d5f39526998067afd66f03aa63df336103/src/object-cache.cls.php#L512

Which calls the PHPRedis extension `del` method.

Which returns `int|false`, 

https://phpredis.github.io/phpredis/Redis.html#method_del

tynanbe:
Fixes #548

### DIFF
--- a/src/object-cache.cls.php
+++ b/src/object-cache.cls.php
@@ -516,7 +516,7 @@ class Object_Cache extends Root {
 			$res = $this->_conn->delete( $key );
 		}
 
-		return $res;
+		return (bool) $res;
 	}
 
 	/**


adrien-robert:
Do you know in which version of LiteSpeed Cache this fix will be include @hi-hai?

PR #556, opened 2023-05-03
Fix generate UCSS for error page

james58899:
Fix sending empty html to cloud causing wrong UCSS to be generated.

Because `self_curl()` will return `false` if it is not 200, and passing it directly to `html_min()` without checking will convert the result to empty html.

https://github.com/litespeedtech/lscache_wp/blob/d1a0aef92424d1c32886f4d594e094ade141d6ad/src/crawler.cls.php#L829-L832

Debug log:


hi-hai:
Good catch. Thanks for the PR.

PR #577, opened 2023-09-17
WPML product category purge

qtwrk:


hi-hai:
Please only provide the change part in PR. The diff has some unrelated format change.

PR #580, opened 2023-09-28
Image Optimisation: Parallel image pull

peter-webbird:
The Image Optimisation process is severely limited by the speed of pulling images down from the QUIC.cloud servers. On websites with thousands of images, this can mean weeks or even months waiting for the image optimisation to complete.

To solve this I have rewritten the pull function to enable parallel pulling of images from remote servers. This significantly improves performance by pulling down multiple images in each pull cron run - by default 15 but it can scale further.

wp_remote_get was replaced with WpOrg\Requests, which is what wp_remote_get uses under the hood, but enables parallel requests. No other major changes were made, but the function had to be significantly reformatted to work asynchronously.

### DIFF
--- a/src/img-optm.cls.php
+++ b/src/img-optm.cls.php
@@ -11,6 +11,9 @@
 
 namespace LiteSpeed;
 
+use WpOrg\Requests\Autoload;
+use WpOrg\Requests\Requests;
+
 defined('WPINC') || exit;
 
 class Img_Optm extends Base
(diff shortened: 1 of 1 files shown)


hi-hai:
Thanks for the PR. Considering the server side load, we may not be able to add this to official release. Do you think its possible to make this a separate plugin and evoke it by a hook? 
image

peter-webbird:
If server side load is a concern, perhaps a better approach is to ship the change with a default value `$imgs_per_req` value of 1 (ie. one request per loop, as it currently is), but allow this to be filtered.

This means that users could easily enable this if they need it (without needing to install a third-party plugin), such as on websites with thousands or tens of thousands of images, but it would be disabled by default so would have minimal impact on server load. An upper limit could also be implemented using the `min` function.

This could work like this:
```suggestion
		// Tune number of images per request for best balance of performance / reliability
		$imgs_per_req = apply_filters( 'litespeed_imgoptm_pull_connection_limit', 1 );
		$imgs_per_req = min( $imgs_per_req, 40 );

		$q = "SELECT * FROM `$this->_table_img_optming` WHERE optm_status = %d ORDER BY id LIMIT %d";
		$_q = $wpdb->prepare($q, array( self::STATUS_NOTIFIED, $imgs_per_req) );
```

Users could then easily filter this using:
```
add_filter( 'litespeed_imgoptm_pull_connection_limit', function( $limit ){ return 10; } );
```

hi-hai:
The idea is brilliant!

I checked the Request lib, it requires PHP5.6. 

I tried to find the introduced date of this lib in WP, didn't find a clear clue. The lib `use WpOrg\Requests\Requests` seems not available in the min WP ver we supported which is WP v4.0.

Do you have any idea on the backward compatibility?

PR #603, opened 2023-12-15
Add direction for Admin Button when Right-To-Left Language

mehrshaddarzi:
When WordPress Language is RTL for example Arabic.
Lightspeed Admin button is not fix style.

please see Screenshot:
without-dir
with-dir


hi-hai:
Applied the compatibility fix in https://github.com/litespeedtech/lscache_wp/commit/50ff4409f46325a3f7bcd241147b0a1f4add1352

PR #608, opened 2023-12-23
Fix an issue where images won't be pulled with older versions of WordPress

tynanbe:
- The `Requests` class didn't always exist at the current namespace.
- The `Autoload` class is a fairly recent addition.

Additionally, the `Requests` class doesn't support all the oldest PHP versions that LSCWP does.

This PR should limit parallel pull to only the most recent versions of WordPress.

Additionally, parallel pull won't be used for PHP < v5.6, as per the Requests [changelog](https://github.com/WordPress/Requests/blob/681bc0553859e650125f562276dbcaf5325eaf31/CHANGELOG.md?plain=1#L123)

### DIFF
--- a/src/img-optm.cls.php
+++ b/src/img-optm.cls.php
@@ -1134,7 +1134,7 @@ class Img_Optm extends Base
 				$total_pulled_ori++;
 			};
 
-			if (class_exists('\WpOrg\Requests\Requests') && class_exists('\WpOrg\Requests\Autoload') && !version_compare(PHP_VERSION, '5.6.0', '<')) {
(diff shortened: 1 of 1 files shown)


src/img-optm.cls.php
```diff
@@ -1124,7 +1134,7 @@ public function pull($manual = false)
 				$total_pulled_ori++;
 			};
 
-			if (class_exists('Requests')) {
+			if (class_exists('\WpOrg\Requests\Requests') && class_exists('\WpOrg\Requests\Autoload') && !version_compare(PHP_VERSION, '5.6.0', '<')) {
```
hi-hai, line 1137: For `!version_compare(PHP_VERSION, '5.6.0', '<')`, why not use `version_compare(PHP_VERSION, '5.6.0', '>=')`

tynanbe: Done~

PR #621, opened 2024-03-01
Cache crawler - run within a certain range of time only

Tymotey:


src/base.cls.php
```diff
@@ -514,6 +515,7 @@ class Base extends Root
 
 		// Crawler
 		self::O_CRAWLER => false,
+		self::O_CRAWLER_SCHEDULE_TIME => '00:00-23:59',
```
hi-hai, line 518: By default it should be empty, which means allow all time.

src/crawler.cls.php
```diff
@@ -255,10 +281,17 @@ public static function async_handler($manually_run = false)
 	 */
 	public static function start($manually_run = false)
 	{
+		$crawler_is_in_time = self::_crawler_in_schedule_time();
```
hi-hai, line 284: This should be done after L289, also considering the context, share the same instance as `self::cls()->_crawl_data($manually_run);`

src/base.cls.php
Tymotey: done

src/crawler.cls.php
Tymotey: Is this better?

src/crawler.cls.php
```diff
@@ -247,6 +247,37 @@ public static function async_handler($manually_run = false)
...
+			foreach ($schedule_times as $time) {
+				if ($time !== '') {
+					$hours = explode('-', $time);
+					$start = new DateTime($hours[0] . ":00");
+					$end = new DateTime($hours[1] . ":00");
+					if ($now < $end && $now > $start) {
+						return true;
+					}
+				}
+			}
```
tynanbe, line 272: This will need more robust parsing and error handling.
What happens for these values of `$time`?
```php
'10';
'1-2';
'100-200';
'1:00:00-2:00:00';
'-2:00';
'00:00-';
'2:00-1:00'; // Should this be corrected or let it stay broken?
'1:00-1:00'; // Nearly impossible for the crawler to run; how to handle it?
'1:59-2:00'; // Valid, but does it cause problems?
// etc.
```

It looks like warnings will be logged in some of these cases. In others, an uncaught `Exception` is thrown.

tynanbe (changes requested):

tynanbe:
@lslisa will need to review wording and document new functionality.

src/crawler.cls.php
Tymotey: Added some parsing to limit incorrect time format

tpl/crawler/settings-general.tpl.php
```diff
@@ -28,6 +28,25 @@
...
 
+		<tr>
+			<th>
+				<?php $id = Base::O_CRAWLER_SCHEDULE_TIME; ?>
+				<?php $this->title($id); ?>
+			</th>
+			<td>
+				<?php $this->build_input($id); ?>
+				<div class="litespeed-desc">
+					<?php echo __('Change the crawler running time.', 'litespeed-cache'); ?>
```
lslisa, line 39: What, specifically, does this refer to? Is it the time the crawler starts running? `running time` implies "how long it runs" so if this setting is for start times, change this to `Specify the time the crawler starts running.`

lslisa: Talked to Tynan about this.
Change to `Specify the crawler running times.`

tpl/crawler/settings-general.tpl.php
```diff
@@ -28,6 +28,25 @@
...
+					<br />
+					<?php echo sprintf( __('You can add multiple times delimited by %s', 'litespeed-cache'), '<code>,</code>'); ?>
```
lslisa, line 41: `You can add multiple ranges in 24-hour format ("HH:mm-HH:mm") delimited by`
(If that time format is incorrect, you can change it. The point is that we should be explicit about what a start-end range looks like)

tpl/crawler/settings-general.tpl.php
Tymotey: @lslisa Changed the text. Please confirm

tpl/crawler/settings-general.tpl.php
Tymotey: @lslisa changed text

src/crawler.cls.php
```diff
@@ -247,6 +249,72 @@ public static function async_handler($manually_run = false)
 		self::start($manually_run);
 	}
 
+	/**
+	 * Check if crawler can run in the choosen time period
+	 *
+	 * @since 6.1
+	 */
+	public static function _crawler_in_schedule_time()
```
hi-hai, line 257: Should this be static?

tpl/crawler/settings-general.tpl.php
```diff
@@ -28,6 +28,25 @@
...
+			<th>
+				<?php $id = Base::O_CRAWLER_SCHEDULE_TIME; ?>
+				<?php $this->title($id); ?>
+			</th>
+			<td>
+				<?php $this->build_input($id); ?>
+				<div class="litespeed-desc">
+					<?php echo __('Specify the crawler running periods.', 'litespeed-cache'); ?>
+					<br />
+					<?php echo sprintf( __('You can add multiple ranges in 24-hour format <code>HH:mm-HH:mm</code> delimited by %s', 'litespeed-cache'), '<code>,</code>'); ?>
```
hi-hai, line 41: Time format shouldn't be in translatable string.

tpl/crawler/settings-general.tpl.php
```diff
@@ -28,6 +28,25 @@
...
+					<br />
+					<?php echo sprintf( __('Server time: %s', 'litespeed-cache'), '<code>'. date('H:m') . '</code>'); ?>
```
hi-hai, line 43: It is redundant to use sprintf here

tpl/crawler/settings-general.tpl.php
```diff
@@ -38,9 +38,9 @@
 				<div class="litespeed-desc">
 					<?php echo __('Specify the crawler running periods.', 'litespeed-cache'); ?>
 					<br />
-					<?php echo sprintf( __('You can add multiple ranges in 24-hour format <code>HH:mm-HH:mm</code> delimited by %s', 'litespeed-cache'), '<code>,</code>'); ?>
+					<?php echo sprintf( __('You can add multiple ranges in 24-hour format <code>%s</code> delimited by %s', 'litespeed-cache'), 'HH:mm-HH:mm', '<code>,</code>'); ?>
```
hi-hai, line 41: Same to <code> tag

tpl/crawler/settings-general.tpl.php
```diff
@@ -38,9 +38,9 @@
...
 					<br />
-					<?php echo sprintf( __('Server time: %s', 'litespeed-cache'), '<code>'. date('H:m') . '</code>'); ?> 
+					<?php echo __('Server time:', 'litespeed-cache') . ' <code>'. date('H:m') . '</code>'; ?>
```
hi-hai, line 43: Same to colon

src/crawler.cls.php
Tymotey: function that calls this function is static

image

tpl/crawler/settings-general.tpl.php
Tymotey: Fixed

tpl/crawler/settings-general.tpl.php
Tymotey: removed and left only __

tpl/crawler/settings-general.tpl.php
Tymotey: Fixed

src/crawler.cls.php
hi-hai: It is different things.

timotei-litespeed: Fixed

PR #626, opened 2024-03-05
Release lane when crawler position is reset

tynanbe:
This should allow the "Manually run" button to work every time as expected.

### DIFF
--- a/src/crawler.cls.php
+++ b/src/crawler.cls.php
@@ -1229,6 +1229,8 @@ class Crawler extends Root
 	{
 		File::save($this->_resetfile, time(), true);
 
+		$this->Release_lane();
+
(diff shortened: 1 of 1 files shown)


hi-hai:
Rather than force release the lane, maybe should check why the crawler stopped?

hi-hai:
Just left some comments for debug direction guidance. 
When reset position is fired, it is supposed to trigger L7373 stopped_reset, then hit _terminate_running, then release.
image
image
If it didn't release, need to find why and maybe some exception we are missing? Shouldn't patch over patch.

hi-hai:
I will close this. Once this issue happens again, @timotei-litespeed plz give it a look.

timotei-litespeed:
ok

PR #630, opened 2024-03-15
Rest class fix some error

timotei-litespeed:
Fix errors log that appear in some cases.
Ticket: https://wordpress.org/support/topic/error-with-litespeed-cache-src-rest-cls-php/

### DIFF
--- a/src/rest.cls.php
+++ b/src/rest.cls.php
@@ -359,7 +359,7 @@ class REST extends Root
 		// Debug2::debug( '[Util] is_rest check [base] ', $rest_url );
 		// Debug2::debug( '[Util] is_rest check [curr] ', $current_url );
 		// Debug2::debug( '[Util] is_rest check [curr2] ', wp_parse_url( add_query_arg( array( ) ) ) );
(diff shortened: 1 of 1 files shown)


src/rest.cls.php
```diff
@@ -359,6 +359,11 @@ public function is_rest($url = false)
 		// Debug2::debug( '[Util] is_rest check [base] ', $rest_url );
 		// Debug2::debug( '[Util] is_rest check [curr] ', $current_url );
 		// Debug2::debug( '[Util] is_rest check [curr2] ', wp_parse_url( add_query_arg( array( ) ) ) );
-		return strpos($current_url['path'], $rest_url['path']) === 0;
+		if($current_url!==false && isset($current_url['path']) && $rest_url!==false && isset($rest_url['path'])){
```
hi-hai, line 362: Checking if its isset or not doesn't help much as using `strpos()` below, should check `empty()` or not.

timotei-litespeed: Done

PR #633, opened 2024-03-19
Fix for ticket #485348 - LSCWP: ESI not working in the Block Editor

timotei-litespeed:
Make sure ESI links are decoded in situation where links are changed by block editor, in render function with wptexturize.

### DIFF
--- a/src/core.cls.php
+++ b/src/core.cls.php
@@ -540,17 +540,19 @@ class Core extends Root
 		$buffer = apply_filters('litespeed_buffer_after', $buffer);
 		
 		// Make sure ESI links are decoded in situation where links are changed by block editor, in render function with wptexturize. Ticket #485348.
-		if(preg_match_all('/(<esi.*src=["\'])(.*)(["\'].*\/>)/mU', $buffer, $matchesEsi)){
+		if(preg_match_all('/(<esi.+src=["\'])(.*)(["\'].*\/>)/mU', $buffer, $matchesEsi)){
 			foreach($matchesEsi[2] as $index => $match){
 				$data = $match;
 				// entity decode needs changes. For example: if finds &#038;
-				if(preg_match('/&#(.*);/mU', $data) === 1){
+				if(preg_match('/&#(.+);/mU', $data) === 1){
 					$data = html_entity_decode($data);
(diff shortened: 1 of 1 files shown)


src/core.cls.php
```diff
@@ -538,6 +538,22 @@ public function send_headers_force($buffer)
 
 		// Hook to modify buffer after
 		$buffer = apply_filters('litespeed_buffer_after', $buffer);
+		
+		// Make sure ESI links are decoded in situation where links are changed by block editor, in render function with wptexturize. Ticket #485348.
+		if(preg_match_all('/(<esi.*src=["\'])(.*)(["\'].*\/>)/mU', $buffer, $matchesEsi)){
```
hi-hai, line 543: `<esi.*src=`
should be
`<esi.+src=`

src/core.cls.php
```diff
@@ -538,6 +538,22 @@ public function send_headers_force($buffer)
...
+			foreach($matchesEsi[2] as $index => $match){
+				$data = $match;
+				// entity decode needs changes. For example: if finds &#038;
+				if(preg_match('/&#(.*);/mU', $data) === 1){
```
hi-hai, line 547: `.*` should be `.+`

src/core.cls.php
```diff
@@ -538,6 +538,22 @@ public function send_headers_force($buffer)
...
+					$data = html_entity_decode($data);
+					$data = rawurldecode($data);
+				}
+				
+				if($data !== $match){
+					$buffer = str_replace($matchesEsi[1][$index].$match.$matchesEsi[3][$index], $matchesEsi[1][$index].$data.$matchesEsi[3][$index], $buffer);
```
hi-hai, line 553: To make it easier to read, set `$matchesEsi[1][$index]` to a var instead of using it twice which makes the row long to read

src/core.cls.php
timotei-litespeed: Changed

src/core.cls.php
timotei-litespeed: Added

src/core.cls.php
timotei-litespeed: changed

PR #635, opened 2024-03-29
ESI - Add Edit with Elementor

timotei-litespeed:


src/esi.cls.php
```diff
@@ -760,7 +763,12 @@ public function sub_widget_block($instance, $widget, $args)
 	public function sub_admin_bar_block()
 	{
 		global $wp_admin_bar;
+		static $rendered = false;
```
hi-hai, line 766: Don't define static here, plz define it in class level. Otherwise others won't be aware of this var and may cause conflict.

src/esi.cls.php
```diff
@@ -771,6 +779,8 @@ public function sub_admin_bar_block()
 		);
 
 		echo $this->sub_esi_block('admin-bar', 'adminbar', $params);
+		
+		$this->admin_rendered = true;
```
hi-hai, line 783: This line is better moved to above line 781. That can give an additional safe check to avoid infinite embed just in case in `$this->sub_esi_block()` it called some hook which fall into this func again. How do you think?

PR #638, opened 2024-04-08
Remove WooCommerce Geolocation Redirect on ESI

timotei-litespeed:
When WooCoommerce Geolocation is active we need to remove the redirection to **?v** query string

### DIFF
--- a/thirdparty/woocommerce.cls.php
+++ b/thirdparty/woocommerce.cls.php
@@ -103,6 +103,13 @@ class WooCommerce extends Base
 				return array_unique($list);
 			});
 		}
+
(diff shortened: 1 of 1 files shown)


thirdparty/woocommerce.cls.php
```diff
@@ -103,6 +103,13 @@ public function add_hooks()
 				return array_unique($list);
 			});
 		}
+
+		// #612331 - remove Woocommerce geolocation redirect on ESI page
+		$link_action = $this->cls('ESI')::QS_ACTION;
```
hi-hai, line 108: will this work in phpv5.3?

PR #641, opened 2024-04-11
Test LTS Cache 6.3  on PHP 5

timotei-litespeed:
Tested admin area and frontend
Tested with wp_debug ON and checked the logs to be empty

### DIFF
--- a/src/debug2.cls.php
+++ b/src/debug2.cls.php
@@ -43,7 +43,8 @@ class Debug2 extends Root
 			!defined('LSCWP_LOG_MORE') && define('LSCWP_LOG_MORE', true);
 		}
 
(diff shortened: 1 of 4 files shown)


src/debug2.cls.php
```diff
@@ -341,7 +342,7 @@ public static function debug($msg, $backtrace_limit = false)
 			return;
 		}
 
-		if (defined('LSCWP_DEBUG_EXC_STRINGS') && Utility::str_hit_array($msg, LSCWP_DEBUG_EXC_STRINGS)) {
+		if (defined('LSCWP_DEBUG_EXC_STRINGS') && Utility::str_hit_array($msg, unserialize(LSCWP_DEBUG_EXC_STRINGS))) {
```
hi-hai, line 345: This is not a good practice. It will cause low efficiency.

PR #644, opened 2024-04-15
Update Instant page to version 5.2.0

timotei-litespeed:
Taken from: https://instant.page/


assets/js/instant_click.ori.js
hi-hai, line 1: Where did you get this original file from? It is not the official original one

PR #647, opened 2024-04-16
HTML - minification blank page

timotei-litespeed:
When large chunk CSS/JS is present on page the page minification is returning empty content. In this case we will return the non minified page and log an error.


src/img-optm.cls.php
```diff
@@ -221,7 +222,7 @@ public function new_req()
 
 		if (!$allowance) {
 			self::debug('❌ No credit');
-			Admin_Display::error(Error::msg('out_of_quota'));
+			Admin_Display::error(Error::msg($err));
```
hi-hai, line 225: This shouldn't be in this PR.

timotei-litespeed: closed and fixed in new PR

PR #651, opened 2024-04-22
Esi nonce in php fix

timotei-litespeed:
When `wp_create_nonce` is assigned in a JSON into html the rendered nonce string contains `<esi>`. 
The fix changes how `wp_verify_nonce` works by adding a check for `<esi`.
If found will replace the tag with the correct nonce value.


hi-hai:
If the PR is still under discussion, please don't mark it as to be merged but draft.

PR #653, opened 2024-04-24
Add DONOTCACHEPAGE condition

timotei-litespeed:
Disable ESI if page is not cached


hi-hai:
image
If you search this const, you can see we have a filter for compatibility concern. To allow manipulation from users, should keep same

timotei-litespeed:
Changed

timotei-litespeed:
👍  Thank you

PR #657, opened 2024-05-01
Rely on just-in-time translation loading

swissspidy:
The plugin was loading translations too early on `plugins_loaded`. Plugins should wait until `init` to ensure the current user has been set up, so that the current user's locale can be respected.

That said, why manually load translations when WordPress can do it for you :-)

Removes the manual `load_plugin_textdomain()` call to rely on WordPress to load translations at the right time, which was added in WordPress 4.6.

Bumps the requirement to WordPress 4.6 accordingly.

### DIFF
--- a/litespeed-cache.php
+++ b/litespeed-cache.php
@@ -178,8 +178,8 @@ if (!function_exists('run_litespeed_cache')) {
 			return;
 		}
 
-		//Check minimum WP requirements, which is 4.0 at the moment.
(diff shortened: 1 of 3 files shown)


szepeviktor:
https://github.com/litespeedtech/lscache_wp/blob/2564cb05b6ba33fbf86fbcb62699dc5d3e28c9ae/litespeed-cache.php#L181-L184 👀

swissspidy:
Ah good catch 😄

PR #664, opened 2024-05-17
Cleanup for litespeed_url table

timotei-litespeed:


timotei-litespeed:
Issue: https://wordpress.org/support/topic/wp_litespeed_url-1-1-gb-in-db-huge-big/

src/data.cls.php
```diff
@@ -465,6 +481,9 @@ public function url_file_clean($file_type)
 		$type = $this->_url_file_types[$file_type];
 		$q = 'DELETE FROM ' . $this->tb('url_file') . ' WHERE `type` = %d';
 		$wpdb->query($wpdb->prepare($q, $type));
+		
+		// Added to cleanup url table. See issue: https://wordpress.org/support/topic/wp_litespeed_url-1-1-gb-in-db-huge-big/
+		$this->table_url_cleanup();
```
hi-hai, line 486: It doesn't have to be a separate func as it handles similar query, for easier context readability, can be just place here.

PR #677, opened 2024-06-11
Cli multisite

timotei-litespeed:
Fix for: CLI "litespeed-purge url" command.

### DIFF
--- a/src/core.cls.php
+++ b/src/core.cls.php
@@ -320,6 +320,9 @@ class Core extends Root
 		switch ($action) {
 			case self::ACTION_QS_PURGE:
 				Purge::set_purge_related();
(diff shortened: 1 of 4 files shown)


src/core.cls.php
```diff
@@ -320,6 +320,9 @@ public function proceed_action($action)
 		switch ($action) {
 			case self::ACTION_QS_PURGE:
 				Purge::set_purge_related();
+				$full_url = (isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] === 'on' ? "https" : "http") . "://$_SERVER[HTTP_HOST]$_SERVER[REQUEST_URI]";
```
hi-hai, line 323: WP has a func `is_ssl()`. Please use WP's func as many as possible

PR #684, opened 2024-06-18
Cloudflare status fix

timotei-litespeed:
If option "litespeed.cdn\cloudflare.status" is emptied(""), if you try to change Cloudflare development status, it will show error.
Ticket #992174


src/cdn/cloudflare.cls.php
```diff
@@ -76,6 +76,10 @@ private function _get_devmode($show_msg = true)
 		Debug2::debug('[Cloudflare] _get_devmode result ', $res);
 
 		$curr_status = self::get_option(self::ITEM_STATUS, array());
+		// Make sure is array: #992174
+		if(!$curr_status){
+			$curr_status = array();
```
hi-hai, line 81: Why not use 
`$curr_status = self::get_option(self::ITEM_STATUS, array()) ?: array();`?

timotei-litespeed: Added

PR #688, opened 2024-06-24
Fix typos

szepeviktor:
Found few misspellings.

@timotei-litespeed This is not a complete PR.
Please add needed changes yourself.


src/base.cls.php
```diff
@@ -110,7 +110,7 @@ class Base extends Root
 	const O_DEBUG_LEVEL = 'debug-level';
 	const O_DEBUG_FILESIZE = 'debug-filesize';
 	const O_DEBUG_COOKIE = 'debug-cookie';
-	const O_DEBUG_COLLAPS_QS = 'debug-collaps_qs';
+	const O_DEBUG_COLLAPSE_QS = 'debug-collapse_qs';
```
hi-hai, line 113: Not sure if const name change will cause fatal error in upgrade in activation part. @Tymotey plz test it.

PR #690, opened 2024-06-27
Cli multisite

timotei-litespeed:
Fixed cli command:
`wp --url=http://test.com/url litespeed-purge url`

When using Multisite and behind LiteSpeed Web ADC

### DIFF
--- a/cli/purge.cls.php
+++ b/cli/purge.cls.php
@@ -157,6 +157,7 @@ class Purge
 	{
 		$data = array(
 			Router::ACTION => Core::ACTION_QS_PURGE,
+			Router::VALIDATE_PURGE => Router::get_hash(Router::VALIDATE_PURGE),
 		);
 		$url = $args[0];
 		$deconstructed = wp_parse_url($url);
@@ -166,6 +167,8 @@ class Purge
 		}
 
 		if (is_multisite()) {
+			$data['switch_blog'] = get_current_blog_id();
+			
 			if (get_blog_id_from_url($deconstructed['host'], '/') === 0) {
 				WP_CLI::error('Multisite url passed in is invalid.');
 				return;
--- a/cli/purge.cls.php
+++ b/cli/purge.cls.php
@@ -157,7 +157,6 @@ class Purge
 	{
 		$data = array(
 			Router::ACTION => Core::ACTION_QS_PURGE,
-			Router::VALIDATE_PURGE => Router::get_hash(Router::VALIDATE_PURGE),
 		);
 		$url = $args[0];
 		$deconstructed = wp_parse_url($url);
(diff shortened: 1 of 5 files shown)


cli/purge.cls.php
```diff
@@ -157,6 +157,7 @@ public function url($args)
 	{
 		$data = array(
 			Router::ACTION => Core::ACTION_QS_PURGE,
+			Router::VALIDATE_PURGE => preg_replace('/.{1,3}(.?)/', '$1', NONCE_KEY),
```
hi-hai, line 160: What did this reg do? Why using this way? Could you please give some comments for code readability?

hi-hai:
Have you tested your change?

timotei-litespeed:
Made changes to the validation:
- tested other actions from CLI and they were failing because of the parameter. Added test to skip validation for Core::ACTION_QS_PURGE
- changed the hash to use Router::get_hash() (saw it in the code and I think it fits better)

timotei-litespeed:
@hi-hai please review the last push "[Router - Get hash improvement](https://github.com/litespeedtech/lscache_wp/pull/690/commits/29c0134c8b92c89c7f049b839be9fa475a0dcb19)" 
Scope of the change is to allow new hashes to be generated and easy to use in future needs

src/router.cls.php
```diff
@@ -501,12 +506,30 @@ private function verify_action()
...
-			// check if it is from admin ip
-			if (!$this->is_admin_ip()) {
+			// check if action is from admin ip. skip test for action Core::ACTION_QS_PURGE.
+			if ( $action != Core::ACTION_QS_PURGE && !$this->is_admin_ip()) {
 				Debug2::debug('[Router] LSCWP_CTRL query string - did not match admin IP: ' . $action);
 				return;
 			}
 
+			$save_blog = get_current_blog_id();
+			if ($_REQUEST['switch_blog']) {
```
hi-hai, line 516: Key may not exist, which will result in warning. Please use `!empty()` to check

src/router.cls.php
```diff
@@ -501,12 +506,30 @@ private function verify_action()
...
+				// If request parameter "switch_blog", switch to correct blog to generate hash.
+				switch_to_blog($_REQUEST['switch_blog']);
+			}
+			$hash = Router::get_hash(self::VALIDATE_PURGE);
+			if ($_REQUEST['switch_blog']) {
+				// Restore blog if needed.
+				switch_to_blog($save_blog);
```
hi-hai, line 523: This switch back and forth is a bit complicated. Should store the hash in the target blog. Please switch at https://github.com/litespeedtech/lscache_wp/pull/690/files#diff-7efe974c01bd97f4667ac581f8ce7fdf45b700734b28b2c945059af3e4b4059aR160

src/router.cls.php
timotei-litespeed: done

cli/purge.cls.php
```diff
@@ -166,7 +166,30 @@ public function url($args)
 		}
 
 		if (is_multisite()) {
-			if (get_blog_id_from_url($deconstructed['host'], '/') === 0) {
+			$current_blog = get_current_blog_id();
+			$path = '/';
+			// If subfolder install test if we can identify the blog from url.
+			// If subdomain install: $path needs to remain '/'.
+			if(!SUBDOMAIN_INSTALL){
```
hi-hai, line 173: Directly calling a const is not safe. Please use `is_subdomain_install()`.
If you want to use const, use like how its used in `is_subdomain_install()` defination.

timotei-litespeed: Changed

cli/purge.cls.php
```diff
@@ -166,7 +166,30 @@ public function url($args)
...
+			if($url_blog && $url_blog !== $current_blog){
+				// If blog found will switch to it.
+				switch_to_blog($url_blog);
+			}
+			else{
+				$path = '/';
+			}
+
+			// Test if link can be found.
+			if (get_blog_id_from_url($deconstructed['host'], $path) === 0) {
```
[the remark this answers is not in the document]
timotei-litespeed: This is used for test if subdomain install has the domain url sent

src/router.cls.php
timotei-litespeed: This was replaced with new code.

src/router.cls.php
```diff
@@ -501,12 +519,20 @@ private function verify_action()
...
 		if (!$this->verify_nonce($action)) {
-			// check if it is from admin ip
-			if (!$this->is_admin_ip()) {
+			// check if action is from admin ip. skip test for action Core::ACTION_QS_PURGE.
+			if ( $action != Core::ACTION_QS_PURGE && !$this->is_admin_ip()) {
 				Debug2::debug('[Router] LSCWP_CTRL query string - did not match admin IP: ' . $action);
 				return;
 			}
 
+			$hash = Router::get_hash(self::VALIDATE_PURGE);
```
hi-hai, line 528: This should be under the cond `$action == Core::ACTION_QS_PURGE`, otherwise it will update hash value for all verify_action()

PR #702, opened 2024-07-25
Php5 lts6.4 fixes

Tymotey:
Tested and fixed issues found with using PHP 5.4


hi-hai:
This will no more needed as we will escalate PHP min ver to v7 since LSCWP v7.0

PR #714, opened 2024-08-02
Delay library improvements

timotei-litespeed:
Rewritten delay fix to improve:
- when JS is run
- how functions and events are run
- improve when DOMContentLoaded is run

For better results of Delay It disable JS Combine


hi-hai:
Will leave this as is as it will require a lot of compatibility tests, and also we will introduce Optimax later which served same purpose.

PR #722, opened 2024-08-14
Update headers in litespeed-cache.php

szepeviktor:
> Since WordPress 5.8 plugin readme files are not parsed for requirements.

> Tested up to: This field ignores minor versions, as plugins shouldn’t break with a minor update.

https://developer.wordpress.org/plugins/wordpress-org/how-your-readme-txt-works/

### DIFF
--- a/litespeed-cache.php
+++ b/litespeed-cache.php
@@ -7,7 +7,6 @@
  * Version:           6.4-a8
  * Requires PHP:      7.2
  * Requires at least: 5.3
- * Tested up to:      6.6
  * Author:            LiteSpeed Technologies
  * Author URI:        https://www.litespeedtech.com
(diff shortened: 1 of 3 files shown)


hi-hai:
Thanks for the PR. Its now in v7.0 branch https://github.com/litespeedtech/lscache_wp/pull/741

szepeviktor:
@hi-hai The other change is based on

> Tested up to: This field ignores minor versions, as plugins shouldn’t break with a minor update.

so it should have only two numbers.

hi-hai:
Thanks for reminding. Fixed in https://github.com/litespeedtech/lscache_wp/commit/cfbb13e2b3f3b8c40d67e91a320683eb2d8d91b1

PR #729, opened 2024-08-23
Updated CSS Load Async to version 3.1.0

timotei-litespeed:
A while ago I looked over the assets versions and this had a new version available
There is not a bug/ticket that have a problem, is an update to the library.
New library version is smaller

Source of update: https://github.com/filamentgroup/loadCSS
Minified with: https://www.toptal.com/developers/javascript-minifier


hi-hai:
Please specify what this update is to fix. @timotei-litespeed

szepeviktor:
This seems like a real solution:
https://web.dev/articles/defer-non-critical-css#optimize

PR #736, opened 2024-09-03
Add elementor nonces

timotei-litespeed:
https://wordpress.org/support/topic/page-caching-including-nonces

### DIFF
--- a/thirdparty/elementor.cls.php
+++ b/thirdparty/elementor.cls.php
@@ -46,4 +46,13 @@ class Elementor
 	{
 		define('LITESPEED_ESI_OFF', true);
 	}
+
+	public static function detect()
+	{
+		/**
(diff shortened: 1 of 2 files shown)


thirdparty/elementor.cls.php
```diff
@@ -46,4 +46,13 @@ public static function disable_litespeed_esi()
 	{
 		define('LITESPEED_ESI_OFF', true);
 	}
+
+	public static function detect()
+	{
```
hi-hai, line 51: @timotei-litespeed This should only run when elementor is enabled

timotei-litespeed: Added

hi-hai:
Please double confirm the branch you committed when marking it as ready in future.

PR #738, opened 2024-09-05
Rely on just-in-time translation loading

swissspidy:
The plugin was loading translations too early on `plugins_loaded`. Plugins should wait until `init` to ensure the current user has been set up, so that the current user's locale can be respected.

That said, why manually load translations when WordPress can do it for you :-)

Removes the manual `load_plugin_textdomain()` call to rely on WordPress to load translations at the right time, which was added in WordPress 4.6.

### DIFF
--- a/src/core.cls.php
+++ b/src/core.cls.php
@@ -81,8 +81,6 @@ class Core extends Root
 		register_uninstall_hook($plugin_file, __NAMESPACE__ . '\Activation::uninstall_litespeed_cache');
 		// }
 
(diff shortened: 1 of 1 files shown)


swissspidy:
Yep 👍 

I noticed they changed this since I opened my previous PR (#657) which I accidentally deleted

szepeviktor:
https://github.com/litespeedtech/lscache_wp/blob/3ec3132c766eb495dbfe519ba9c76695372e8b9e/litespeed-cache.php#L183

PR #740, opened 2024-09-06
Ci actions and prettier change

timotei-litespeed:
Updates to Prettier files it format/check
Update PHP version for Prettier. Will sync with LSC v7


timotei-litespeed:
Related to this: https://github.com/litespeedtech/lscache_wp/pull/742 which is improving this PR

PR #741, opened 2024-09-09
Add min php version

timotei-litespeed:
https://developer.wordpress.org/plugins/wordpress-org/how-your-readme-txt-works/


szepeviktor:
Very similar to #722

timotei-litespeed:
@szepeviktor Indeed :) We will look over the PR and modify them into 7.0

PR #747, opened 2024-09-20
Remove deprecated notice

timotei-litespeed:
Removing deprecated notice from this thread: https://wordpress.org/support/topic/php-deprecated-notice-14/

The fix is made to catch a case where the url cannot be parsed and returning null.

### DIFF
--- a/src/cdn.cls.php
+++ b/src/cdn.cls.php
@@ -312,7 +312,9 @@ class CDN extends Root
 
 			// Parse file postfix
 			$parsed_url = parse_url($url, PHP_URL_PATH);
-			if(is_null($parsed_url)) continue;
+			if (!$parsed_url) {
+				continue;
+			}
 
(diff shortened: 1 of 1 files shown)


src/cdn.cls.php
```diff
@@ -311,7 +311,10 @@ private function _replace_inline_css()
 			$url = str_replace(array(' ', '\t', '\n', '\r', '\0', '\x0B', '"', "'", '&quot;', '&#039;'), '', $url);
 
 			// Parse file postfix
-			$postfix = '.' . pathinfo(parse_url($url, PHP_URL_PATH), PATHINFO_EXTENSION);
+			$parsed_url = parse_url($url, PHP_URL_PATH);
+			if(is_null($parsed_url)) continue;
```
hi-hai, line 315: maybe you can just check `if(!$parse_url)`?

timotei-litespeed: Fixed

hi-hai:
Plz contact the user to confirm its fixed.

PR #748, opened 2024-09-20
Fix modal footer

timotei-litespeed:
Fix to allow PHP CS to finish running


hi-hai:
Seems this one includes unrelated commits as well. May be related to my force push. Can you remake it?

PR #759, opened 2024-10-17
Add PHP info to report

timotei-litespeed:
Added checkbox that will add PHP info to report

Also fixed a warning on report.cls.php:153 where array key is undefined.

### DIFF
--- a/tpl/toolbox/report.tpl.php
+++ b/tpl/toolbox/report.tpl.php
@@ -76,7 +76,7 @@ $install_link = Utility::build_url(Router::ACTION_ACTIVATION, Activation::TYPE_I
 								'litespeed-cache'
 							),
 							'<code>INFO_GENERAL</code>, <code>INFO_CONFIGURATION</code> and <code>INFO_MODULES</code>',
-							'<a href="https://www.php.net/manual/en/function.phpinfo.php" traget="__blank">phpinfo()</a>'
+							'<a href="https://www.php.net/manual/en/function.phpinfo.php" target="__blank">phpinfo()</a>'
 						),
 						false
 					);
--- a/tpl/toolbox/report.tpl.php
+++ b/tpl/toolbox/report.tpl.php
@@ -65,6 +65,25 @@ $install_link = Utility::build_url(Router::ACTION_ACTIVATION, Activation::TYPE_I
 				</td>
 			</tr>
 			<tr>
+				<th>&nbsp;</th>
+				<td>
+					<?php
+					$this->build_checkbox(
+						'attach_php',
+						sprintf(
+							__(
+								'Attach PHP info to report. It will use data returned by function %1$s, sections %2$s.',
+								'litespeed-cache'
+							),
+							'<a href="https://www.php.net/manual/en/function.phpinfo.php" traget="__blank">phpinfo()</a>',
+							'<code>INFO_GENERAL, INFO_CONFIGURATION, INFO_MODULES</code>'
+						),
+						false
+					);
+					?>
+				</td>
+			</tr>
+			<tr>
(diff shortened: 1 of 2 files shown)


tpl/toolbox/report.tpl.php
```diff
@@ -64,6 +64,21 @@
...
 				</td>
 			</tr>
+			<tr>
+				<th>&nbsp;</th>
+				<td>
+					<?php
+					$this->build_checkbox(
+						'attach_php',
+						sprintf(
+							__('Attach PHP info to report. It will use data returned by function <a %s>phpinfo()</a>, sections <code>INFO_GENERAL, INFO_CONFIGURATION, INFO_MODULES</code>.', 'litespeed-cache'),
```
hi-hai, line 74: Did you get Lisa proof reading? 
Besides, all the code/func related words shouldn't be in translation. Move the whole part into %s plz. If there are more than 1 placeholder, use %2$s to avoid disorder in diff langs.

timotei-litespeed: Changed

timotei-litespeed:
@lslisa Need help! :) Please proofread the text added :)
Thank you!

lslisa:
```
Attach PHP info to report. Check this box to insert the <code>INFO_GENERAL, INFO_CONFIGURATION, INFO_MODULES</code> <a %s>phpinfo()</a> sections.
```
(Please double-check that I didn't inadvertently mess up any of the syntax when I moved things around. I think it's still ok.)

timotei-litespeed:
@lslisa Thank you! Changes have been applied

tpl/toolbox/report.tpl.php
```diff
@@ -64,6 +64,25 @@
...
+				<td>
+					<?php
+					$this->build_checkbox(
+						'attach_php',
+						sprintf(
+							__(
+								'Attach PHP info to report. Check this box to insert the %1$s sections from %2$s.',
+								'litespeed-cache'
+							),
+							'<code>INFO_GENERAL</code>, <code>INFO_CONFIGURATION</code> and <code>INFO_MODULES</code>',
```
hi-hai, line 78: `and` shouldn't be out of translation. Plz join these 3 w/ `/`.

lslisa: Grammatically, I'm not a fan of `/`. Can you do something like this? (not sure about the syntax)
```
	'Attach PHP info to report. Check this box to insert the %1$s and %2$s sections from %3$s.',
								'litespeed-cache'
							),
							'<code>INFO_GENERAL</code>, <code>INFO_CONFIGURATION</code>',
							'<code>INFO_MODULES</code>',
							'<a href="https://www.php.net/manual/en/function.phpinfo.php" target="__blank">phpinfo()</a>'
```
(BTW, you had `target` misspelled as `traget` in the last line)

timotei-litespeed: @hi-hai 
I thought to keep them in translation, but if we change the parameters(add/remove flags) we will not need to change text translation

the split was @lslisa  ideea. I am ok changing them  to 1 code and /  between, but the change was suggested by her

lslisa: If we can avoid using `/` and keep the `and` in there by using an additional parameter like I suggested, that would be nicer.

hi-hai: Those consts explanation are unnecessary to the end users. Maybe we can just remove them and simply say phpinfo?

timotei-litespeed: @hi-hai like this?

lslisa:
I know I'm late to the party here, since this has already been merged, but grammatically it would be better without a `the` before `data`. I would change it to say `Check this box to insert relevant data from`

PR #761, opened 2024-10-21
Check filename is valid or not

AkramiPro:
in some cases, filename is null and fire PHP warning on `direname` function.

### DIFF
--- a/litespeed-cache.php
+++ b/litespeed-cache.php
@@ -4,7 +4,7 @@
  * Plugin Name:       LiteSpeed Cache
  * Plugin URI:        https://www.litespeedtech.com/products/cache-plugins/wordpress-acceleration
  * Description:       High-performance page caching and site optimization from LiteSpeed
- * Version:           6.5.1
+ * Version:           6.5.2
  * Author:            LiteSpeed Technologies
  * Author URI:        https://www.litespeedtech.com
  * License:           GPLv3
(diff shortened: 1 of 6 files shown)


timotei-litespeed:
Thank you for the fix. We will review it and decide what will happen to it.

One thing that will help us: is to give details of when the warning appears. What action(s) you do for the warning to appear.

Tim

AkramiPro:
I should mention that this is just a quick fix for suppress PHP warning.
in fact, we have no debug functionality when using object cache and it must be fixed.
`Debug2` class needs to be changed to support headless running without any dependency on core.

timotei-litespeed:
You are right about just hidding it, that's why I asked you about the error.
What is the version of the plugin you see the warning?

AkramiPro:
set this in wp-config.php
```
define( 'LSCWP_LOG', true );
```

timotei-litespeed:
Ok, that makes sense :)
LSCWP_LOG is not made to be used like that.
Why do you added the LSCWP_LOG in wp-config?

I will look over the Debug class to remove the warning in this case.
Thank you for your details! :)

AkramiPro:
yes, you are right. i look at the source and to force object cache create logs make some changes and forgot that  LSCWP_LOG  must not set manually.
anyway, check for filename is always a good practice :)

timotei-litespeed:
@hi-hai 
As per above findings our code will not throw error.
There is the possibility of user defining `define( 'LSCWP_LOG', true );` in wp-config.php and that will show errors when object cache is activated.
PR goal is to remove the warnings that will appear when dirname from `File::save()` will get incorrect filename.

timotei-litespeed:
@AkramiPro thank you for the informations and we will further decide what is the correct action on the situation you described

hi-hai:
> yes, you are right. i look at the source and to force object cache create logs make some changes and forgot that LSCWP_LOG must not set manually. anyway, check for filename is always a good practice :)

That makes sense. Added in https://github.com/litespeedtech/lscache_wp/commit/6e79c5c919a789dce6c46294934707d389dfce2a. Thanks.

timotei-litespeed:
@AkramiPro 👍

PR #767, opened 2024-10-31
Purge actions and filter

timotei-litespeed:
Add filters for post purge events
Add actions for each major purge event

Changes are done to help users add their own actions and functionalities.
Example most recent: [https://wordpress.org/support/topic/delete-entry-cache-when-saving/#post-18106071](https://wordpress.org/support/topic/delete-entry-cache-when-saving/#post-18106071)

### DIFF
--- a/src/purge.cls.php
+++ b/src/purge.cls.php
@@ -51,14 +51,17 @@ class Purge extends Base
 	public function init()
 	{
 		//register purge actions
-		$purge_post_events = array(
-			// 'edit_post',
-			// 'save_post',
-			'delete_post',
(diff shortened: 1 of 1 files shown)


hi-hai:
The filter part is fine. Why do we need to add those many actions? @timotei-litespeed

timotei-litespeed:
@hi-hai to allow users add custom code to run after different purge actions

hi-hai:
Can u use dev to commit again? We froze v7.0 branch already

timotei-litespeed:
done #776

PR #769, opened 2024-11-06
style adj.

kokers:
unclosed div fix, diff link if QC CDN enabled, mobile fixes


hi-hai:
Moved to https://github.com/litespeedtech/lscache_wp/pull/773

PR #774, opened 2024-11-07
Update dashboard code to limit Page Load time just for LTS servers

timotei-litespeed:


hi-hai:
@timotei-litespeed Plz resolve the conflict

PR #776, opened 2024-11-08
PR Redo - Purge actions and filter

timotei-litespeed:
Add filters for post purge events
Add actions for each major purge event + parameters to some actions

### DIFF
--- a/src/purge.cls.php
+++ b/src/purge.cls.php
@@ -50,17 +50,14 @@ class Purge extends Base
 	 */
 	public function init()
 	{
-		//register purge actions
-		// Mostly used values: edit_post, save_post, delete_post, wp_trash_post, clean_post_cache, wp_update_comment_count
-		$purge_post_events = add_filter(
-			'litespeed_purge_post_events',
-			array(
-				'delete_post',
-				'wp_trash_post',
-				// 'clean_post_cache', // This will disable wc's not purge product when stock status not change setting
-				'wp_update_comment_count', // TODO: check if needed for non ESI
-			)
-		);
+		// Register purge actions.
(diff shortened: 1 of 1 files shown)


src/purge.cls.php
```diff
@@ -797,6 +810,7 @@ public function purge_url($url, $purge2 = false, $quite = false)
 		self::add($hash, $purge2);
 
 		!$quite && !defined('LITESPEED_PURGE_SILENT') && Admin_Display::success(sprintf(__('Purge url %s', 'litespeed-cache'), $val));
+		do_action('litespeed_purged_link');
```
hi-hai, line 813: @timotei-litespeed Maybe add the url as param?

timotei-litespeed: Added and looked over other functions and made a few changes

src/purge.cls.php
```diff
@@ -924,11 +945,14 @@ public static function purge_comment_widget()
...
 			return;
 		}
+
 		$recent_comments = $wp_widget_factory->widgets['WP_Widget_Recent_Comments'];
 		if (!is_null($recent_comments)) {
 			self::add(Tag::TYPE_WIDGET . $recent_comments->id);
 			self::add_private(Tag::TYPE_WIDGET . $recent_comments->id);
 		}
+
+		do_action('litespeed_purged_comment_widget', $recent_comments->id);
```
hi-hai, line 955: This will result in error if you don't use the way L950 does.

src/purge.cls.php
hi-hai: Good try. There are still some places you may miss. Also some comments added

PR #778, opened 2024-11-12
Fix warning in Buddy Press code integration

timotei-litespeed:
Based on the topic: [https://wordpress.org/support/topic/undefined-key-array-warning/#post-18129398](https://wordpress.org/support/topic/undefined-key-array-warning/#post-18129398)

### DIFF
--- a/thirdparty/bbpress.cls.php
+++ b/thirdparty/bbpress.cls.php
@@ -78,10 +78,10 @@ class BBPress
 		}
 
 		global $wp_widget_factory;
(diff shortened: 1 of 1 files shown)


szepeviktor:
Maybe using one these methods? https://developer.wordpress.org/reference/classes/wp_widget_factory/

hi-hai:
> Maybe using one of these methods? https://developer.wordpress.org/reference/classes/wp_widget_factory/

Good point. @timotei-litespeed plz improve your PR accordingly.

timotei-litespeed:
Changed

PR #781, opened 2024-11-26
Fix notice and wrong Regex

timotei-litespeed:
Found after looking for this issue: https://wordpress.org/support/topic/php-deprecated-notice-14/

### DIFF
--- a/src/cdn.cls.php
+++ b/src/cdn.cls.php
@@ -306,8 +306,12 @@ class CDN extends Root
 		 * @see  #685485
 		 * @since 3.0
 		 */
-		preg_match_all('/url\((?![\'"]?data)[\'"]?([^\)\'"\\\]+)[\'"]?\)/i', $this->content, $matches);
+		preg_match_all('/url\((?![\'"]?data)[\'"]?(.*?)[\'"]?\)/i', $this->content, $matches);
 		foreach ($matches[1] as $k => $url) {
+			if ($url == null) {
+				continue;
+			}
+
 			$url = str_replace(array(' ', '\t', '\n', '\r', '\0', '\x0B', '"', "'", '&quot;', '&#039;'), '', $url);
 
 			// Parse file postfix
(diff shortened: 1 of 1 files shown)


src/cdn.cls.php
```diff
@@ -306,8 +306,12 @@ private function _replace_inline_css()
 		 * @see  #685485
 		 * @since 3.0
 		 */
-		preg_match_all('/url\((?![\'"]?data)[\'"]?([^\)\'"\\\]+)[\'"]?\)/i', $this->content, $matches);
+		preg_match_all('/url\((?![\'"]?data)[\'"]?(.*?)[\'"]?\)/i', $this->content, $matches);
 		foreach ($matches[1] as $k => $url) {
+			if ($url == null) {
```
hi-hai, line 311: Should we just use `if (!$url)` as empty case should be bypassed as well??

timotei-litespeed: Yes, added the changed

src/cdn.cls.php
```diff
@@ -306,8 +306,12 @@ private function _replace_inline_css()
 		 * @see  #685485
 		 * @since 3.0
 		 */
-		preg_match_all('/url\((?![\'"]?data)[\'"]?([^\)\'"\\\]+)[\'"]?\)/i', $this->content, $matches);
+		preg_match_all('/url\((?![\'"]?data)[\'"]?(.*?)[\'"]?\)/i', $this->content, $matches);
```
hi-hai, line 309: Should we change `(.*?)` to `(.+?)`? @timotei-litespeed

src/cdn.cls.php
```diff
@@ -306,8 +306,12 @@ private function _replace_inline_css()
 		 * @see  #685485
 		 * @since 3.0
 		 */
-		preg_match_all('/url\((?![\'"]?data)[\'"]?([^\)\'"\\\]+)[\'"]?\)/i', $this->content, $matches);
+		preg_match_all('/url\((?![\'"]?data)[\'"]?(.+?)[\'"]?\)/i', $this->content, $matches);
 		foreach ($matches[1] as $k => $url) {
+			if (!$url) {
```
hi-hai, line 311: As its used + now, is this condition redundant? @timotei-litespeed

timotei-litespeed: removed

PR #793, opened 2025-01-17
Fix CS

szepeviktor:
This should make CI happy.


szepeviktor:
@hi-hai ✔️

tpl/inc/metabox.php
```diff
@@ -18,7 +18,13 @@
...
 	} else {
 		echo '<div style="display:flex;margin-bottom:10px;align-items: center;gap: 2ch;justify-content: space-between;"><label for="' . $k . '">' . $v . '</label>';
-		echo '<input class="litespeed-tiny-toggle" id="' . Str::trim_quotes($k) . '" name="' . Str::trim_quotes($k) . '" type="checkbox" value="1" ' . ($existing_val ? 'checked' : '') . ' />';
+		echo '<input class="litespeed-tiny-toggle" id="' .
+			Str::trim_quotes($k) .
+			'" name="' .
+			Str::trim_quotes($k) .
+			'" type="checkbox" value="1" ' .
+			($existing_val ? 'checked' : '') .
+			' />';
```
szepeviktor, line 27: With a single class you could say
```php
$input_attrs = [
    'class' => 'litespeed-tiny-toggle',
    'id' => Str::trim_quotes($k),
    'name' => Str::trim_quotes($k),
    'type' => 'checkbox',
    'value' => '1',
];
if ($existing_val) {
    $input_attrs['checked'] = null;
}
echo (new Element('input', $input_attrs))->render();
```
https://github.com/szepeviktor/SentencePress/blob/master/src/Html/Element.php 👀

hi-hai: That is a good idea, just a bit overkill in current context maybe? 😌

hi-hai:
Thanks for the PR. Kindly remind: the PR needs to take `dev` branch while not `master`.

tpl/inc/metabox.php
hi-hai: True. A sprintf could save you from that 😆

szepeviktor: I've added a new method only for you.
https://github.com/szepeviktor/SentencePress/commit/4845a8168529c1a0d8370c3186dc2b34f777bddc
so you can
```php
$input = (new Element('input', $input_attrs))->setBooleanWhen('checked', $existing_val);
echo $input->render();
```

PR #794, opened 2025-01-17
Image Optimization - Fix Auto Pull

timotei-litespeed:


src/task.cls.php
```diff
@@ -44,7 +44,7 @@ public function init()
 		$guest_optm = $this->conf(Base::O_GUEST) && $this->conf(Base::O_GUEST_OPTM);
 
 		foreach (self::$_triggers as $id => $trigger) {
-			if (!$this->conf($id)) {
+			if (!$this->conf($id) && $id !== 'img_optm-cron' ) {  // @Deprecated since v7.0 TODO: remove after v7.5
```
hi-hai, line 47: Use `Base::O_IMG_OPTM_CRON` instead of hardcoded string

timotei-litespeed: Yes!

timotei-litespeed: Yes, you are right :)

PR #795, opened 2025-01-21
Fix spamming object cache connection messages

timotei-litespeed:
Moved Object Cache messages to show only if "Debug Log" is set to "Admin IP Only"
Removed extra debug messages
On connection: I combined 2 lines into 1 message


src/object-cache.cls.php
```diff
@@ -154,7 +152,7 @@ private function debug_oc($text, $show_error = false)
 			return;
 		}
 
-		if (!$show_error && !$this->_cfg_debug) {
+		if (!$show_error && $this->_cfg_debug!=2) {
```
hi-hai, line 155: We don't use 2
image

PR #796, opened 2025-01-23
Typo breaks fallback constant

Roy-Orbison:
Optimisation bypass is ignored.


szepeviktor:
Static analysis can catch these easily.
Please consider using PHPStan.

PR #798, opened 2025-01-24
Use built-in to get random array value

Roy-Orbison:
`array_rand()` has been around since PHP4 and already uses Mersenne Twister, internally.


Roy-Orbison:
This PR is not causing the CI failure.

PR #800, opened 2025-01-30
CDN - remove filetypes to replace if setting is disabled

timotei-litespeed:
In CDN, at mapping settings if Include Images/CSS/JS is disabled it removes the file types from "Include File Types" setting.
Before:
image
After save:
image

### DIFF
--- a/src/admin-settings.cls.php
+++ b/src/admin-settings.cls.php
@@ -117,6 +117,35 @@ class Admin_Settings extends Base
 						foreach ($data as $k => $v) {
 							if ($child == self::CDN_MAPPING_FILETYPE) {
 								$v = Utility::sanitize_lines($v);
+
+								// Remove from MAPPING FILETYPE IMAGES/CSS/JS if settings are disabled
+								$remove_type = array();
+								if (empty($data2[$k][self::CDN_MAPPING_INC_IMG])) {
+									$remove_type = array_merge($remove_type, array('jpg', 'jpeg', 'png', 'gif', 'svg', 'webp', 'avif'));
+								}
+								if (empty($data2[$k][self::CDN_MAPPING_INC_CSS])) {
+									$remove_type[] = 'css';
+								}
+								if (empty($data2[$k][self::CDN_MAPPING_INC_JS])) {
+									$remove_type[] = 'js';
+								}
+
+								if (count($remove_type) > 0) {
+									$temp = array_filter($v, function ($i_v) use ($remove_type) {
+										$leave_value = true;
+
+										foreach ($remove_type as $remove) {
+											if (strpos($i_v, $remove) !== false) {
+												$leave_value = false;
+												break;
+											}
+										}
+
+										return $leave_value;
+									});
+
+									$v = array_values($temp);
+								}
 							}
 							if ($child == self::CDN_MAPPING_URL) {
 								# If not a valid URL, turn off CDN


src/admin-settings.cls.php
```diff
@@ -117,6 +117,35 @@ public function save($raw_data)
 						foreach ($data as $k => $v) {
 							if ($child == self::CDN_MAPPING_FILETYPE) {
 								$v = Utility::sanitize_lines($v);
+
+								// Remove from MAPPING FILETYPE IMAGES/CSS/JS if settings are disabled
+								$remove_type = array();
+								if (empty($data2[$k][self::CDN_MAPPING_INC_IMG])) {
+									$remove_type = array_merge($remove_type, array('jpg', 'jpeg', 'png', 'gif', 'svg', 'webp', 'avif'));
```
hi-hai, line 124: These types should be always removed, no matter if the option checked or not. Append them dynamically in `cdn::init()`

timotei-litespeed: Removed the conditions

src/admin-settings.cls.php
```diff
@@ -117,6 +117,22 @@ public function save($raw_data)
 						foreach ($data as $k => $v) {
 							if ($child == self::CDN_MAPPING_FILETYPE) {
 								$v = Utility::sanitize_lines($v);
+
+								// Remove from MAPPING FILETYPE extensions for IMAGES, CSS, JS
+								$remove_type = array('jpg', 'jpeg', 'png', 'gif', 'svg', 'webp', 'avif', 'css', 'js');
```
hi-hai, line 122: give it a filter plz

src/admin-settings.cls.php
```diff
@@ -117,6 +117,22 @@ public function save($raw_data)
...
+								$temp = array_filter($v, function ($i_v) use ($remove_type) {
```
hi-hai, line 123: just a simple loop to unset it, easier to read.

src/admin-settings.cls.php
```diff
@@ -117,6 +117,22 @@ public function save($raw_data)
...
+									$leave_value = true;
+
+									foreach ($remove_type as $remove) {
+										if (strpos($i_v, $remove) !== false) {
```
hi-hai, line 127: this is not correct. Need to full match

src/cdn.cls.php
```diff
@@ -209,9 +209,28 @@ private function _finalize()
 			$this->_replace_inline_css();
 		}
 
+		// Temporary add file types to mapping filetype rewrite
```
hi-hai, line 212: I think you should append them in `init()`

src/admin-settings.cls.php
timotei-litespeed: fixed

src/cdn.cls.php
timotei-litespeed: Moved

src/admin-settings.cls.php
timotei-litespeed: fixed with above

src/admin-settings.cls.php
```diff
@@ -117,6 +117,20 @@ public function save($raw_data)
...
+									'.jpeg',
+									'.png',
+									'.gif',
+									'.svg',
+									'.webp',
+									'.avif',
+									'.css',
+									'.js',
+								));
+								$v = array_diff($v, $remove_type);
```
hi-hai, line 133: 👍

timotei-litespeed:
@hi-hai fixed the file types to include images.
Also removed extra code that I wrote and was not necessary.
Tested .attribute + element.attribute  both cases.

PR #801, opened 2025-02-04
Add Cookie path check

timotei-litespeed:
Based on this report https://wordpress.org/support/topic/fatal-error-in-litespeed-cache-setcookie-path-issue/
Added the code change from this PR.

### DIFF
--- a/src/vary.cls.php
+++ b/src/vary.cls.php
@@ -731,6 +731,19 @@ class Vary extends Root
 		return false;
 	}
 
+	private function test_cookie_path($path)
+	{
+		if ($path[0] != '/') {
+			return false;
+		}
+
+		if (preg_match('/[\x00-\x1F\x7F]/', $path)) {
+			return false;
+		}
+
+		return true;
+	}
+
 	/**
 	 * Set the vary cookie.
 	 *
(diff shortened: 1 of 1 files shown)


src/vary.cls.php
```diff
@@ -731,6 +731,19 @@ private function _get_cookie_val($key)
 		return false;
 	}
 
+	private function test_cookie_path($path)
+	{
+		if ($path[0] != '/') {
```
szepeviktor, line 736: This may function better.
```suggestion
		if (substr($path, 0, 1) !== '/') {
```

timotei-litespeed: fixed. thank you! :)

hi-hai:
@timotei-litespeed this is not a valid fix. It will invalidate the func, defeat the setcookie.

You should locate the issue first. Figure out what char caused the error. As commented in the src code, `// only set commenter status for current domain path` this is the reason we set the cookie. So, should NOT bypass the path.

src/vary.cls.php
```diff
@@ -754,6 +754,9 @@ private function _cookie($val = false, $expire = false, $path = false)
 		 */
 		$is_ssl = $this->conf(Base::O_UTIL_NO_HTTPS_VARY) ? false : is_ssl();
 
-		setcookie(self::$_vary_name, $val, $expire, $path ?: COOKIEPATH, COOKIE_DOMAIN, $is_ssl, true);
+		// Fix https://wordpress.org/support/topic/fatal-error-in-litespeed-cache-setcookie-path-issue/
+		$encoded_path = urlencode( $path ?: COOKIEPATH, COOKIE_DOMAIN );
```
szepeviktor, line 758:
```suggestion
		$encoded_path = urlencode( $path ?: COOKIEPATH);
```
`urlencode` has only one parameter.
https://www.php.net/manual/en/function.urlencode.php

hi-hai:
@timotei-litespeed Please work w/ user to locate the issue, before PR. Once user confirmed fix, give a new one.
I will close this PR.

PR #805, opened 2025-02-11
Crawler - Use alternate port for crawler when default is not open

timotei-litespeed:
Fix crawler port if default is not open.
Fix a code format error.

### DIFF
--- a/src/crawler.cls.php
+++ b/src/crawler.cls.php
@@ -784,6 +784,31 @@ class Crawler extends Root
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


hi-hai:
Done in https://github.com/litespeedtech/lscache_wp/commit/88153346236799b5c79ec7e57e12e7b90bc6c630#diff-53cf86e139cba33fedb601ef890f35c7f575cc8935cfb2005cea05ea6a509953R672

PR #811, opened 2025-02-18
Htaccess change vary list

timotei-litespeed:
Remove extra comma from vary list


hi-hai:
Are you sure it should be removed?

PR #812, opened 2025-02-19
OPcache - handle opcache.restrict_api setting

timotei-litespeed:
When opcache parameter **restrict_api** is set there is a warning when OPCache is cleared.
Reading: [https://www.php.net/manual/en/opcache.configuration.php#ini.opcache.restrict-api](https://www.php.net/manual/en/opcache.configuration.php#ini.opcache.restrict-api) I see that it also blocks the request.

Fix: if **restrict_api** is not set or empty, reset cache will work.
If path from **restrict_api** is found in script that starts the reset, it will work.
Eg: **opcache.restrict_api="/home/ssssss.com/public_html"** 

Tickets:
https://wordpress.org/support/topic/zend-opcache-api-is-restricted-by-restrict_api-5/
#9496550

### DIFF
--- a/src/purge.cls.php
+++ b/src/purge.cls.php
@@ -398,11 +398,34 @@ class Purge extends Base
 	public function purge_all_opcache($silence = false)
 	{
 		if (!Router::opcache_enabled()) {
-			self::debug('Failed to reset opcode cache due to opcache not enabled');
+			self::debug('❌ Failed to reset opcode cache due to opcache not enabled');
 
 			if (!$silence) {
 				$msg = __('Opcode cache is not enabled.', 'litespeed-cache');
-				Admin_Display::error($msg);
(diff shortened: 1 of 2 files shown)


src/purge.cls.php
```diff
@@ -393,16 +393,40 @@ private function _purge_all_cssjs($silence = false)
...
 	 *
 	 * @since  1.8.2
+	 * @since  7.3 added test for opcode cache restriction
 	 * @access public
 	 */
 	public function purge_all_opcache($silence = false)
 	{
 		if (!Router::opcache_enabled()) {
-			self::debug('Failed to reset opcode cache due to opcache not enabled');
+			self::debug('❌ Failed to reset opcode cache due to opcache not enabled');
```
lslisa, line 402: Failed to reset OPcode cache due to OPcache not enabled.

PR #819, opened 2025-03-17
Purge - Make sure the event list contains unique values

timotei-litespeed:


hi-hai:
Do we need this? Any case it will get into duplicate values?

timotei-litespeed:
if multiple code will add(for example:  **save_post** ) it will appear many times

hi-hai:
@timotei-litespeed I dont think we need this. Anyone who use the filter needs to make sure they don't duplicate. Its a basic coding skill to check existence of key instead of blindly add.

hi-hai:
If do here, you need to do everywhere. Does that make sense?

timotei-litespeed:
Got it :) I see the reason :)

PR #825, opened 2025-04-07
Activation - Fix htaccess generation

timotei-litespeed:
Test case:
LSC is active
_.htaccess_ file has the correct content
Deactivate LSC
_.htaccess_ has LSC data removed
Activate LSC
_.htaccess_ will not have LSC data
Save LSC settings
_.htaccess_ will have LSC data required

### DIFF
--- a/src/activation.cls.php
+++ b/src/activation.cls.php
@@ -72,9 +72,8 @@ class Activation extends Base
 					}
 				}
 			}
-		} else {
-			self::cls('Activation')->update_files();
 		}
+		self::cls()->update_files();
 
 		if (defined('LSCWP_REF') && LSCWP_REF == 'whm') {
 			GUI::update_option(GUI::WHM_MSG, GUI::WHM_MSG_VAL);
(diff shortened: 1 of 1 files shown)


timotei-litespeed:
@hi-hai thank you

src/activation.cls.php
```diff
@@ -74,6 +72,8 @@ public static function register_activation()
 					}
 				}
 			}
+		} else {
+			self::cls('Activation')->update_files();
```
hi-hai, line 76: Why not `$this->update_files();`?

hi-hai: @timotei-litespeed

timotei-litespeed: @hi-hai 
I read the line 67 comment and I thought it requred to run only in single site
But after testing on multi site: I removed the _else_ and it works.
Fixed

timotei-litespeed: @hi-hai added the improvement we spoke

PR #832, opened 2025-04-18
Tools - update API check

timotei-litespeed:
I tested why the IP is empty and I found out the integration is returning IPV6. I added a debug log to show if IP is not valid(IPV4).

I looked for docs on **IP.me** but I was not able to find any.

Searched for alternatives and I found: [https://www.ipify.org/](https://www.ipify.org/) which has API's dedicated to each version of IP. I changed the code to use V4 only.


hi-hai:
Will not use this site.

PR #834, opened 2025-04-22
Replace use of `SHOW TABLES` with `DESCRIBE`.

boonebgorges:
I recently updated from 6.x to 7.0.1 on a very large WP Multisite installation. MySQL ground to a halt shortly afterward, and after investigation we determined that the enormous number of `SHOW TABLES` queries coming from `litespeed-cache` were the culprit.

`SHOW TABLES` requires touching `INFORMATION_SCHEMA`, which on very large databases means temp sort and other operations that require enormous resources. See https://bugs.mysql.com/bug.php?id=60961, https://www.percona.com/blog/solving-information_schema-slowness/, https://forums.mysql.com/read.php?24,706979,706979

Every instance of `SHOW TABLES` in this plugin uses a hardcoded value (no wildcards) and is only check for existence. As such, we can use `DESCRIBE` (or alternatively `SHOW CREATE TABLE`), which only loads information about the queried table and doesn't require loading all table info into memory.

I'm running this patch on my client site and the performance problems have gone away. 

Thanks in advance for considering.


timotei-litespeed:
@boonebgorges We decided to go further with all the changes and created a new PR: #850

You can test the PR from `LSC` => `Toolbox` => `Beta Test` ([https://docs.litespeedtech.com/lscache/lscwp/toolbox/#beta-test-tab](https://docs.litespeedtech.com/lscache/lscwp/toolbox/#beta-test-tab)) entering the latest commit `https://github.com/timotei-litespeed/lscache_wp/commit/ab3bd000ea55b196604a94aa8a1dab669382a073` in field.

Thank you  :)

hi-hai:
https://github.com/litespeedtech/lscache_wp/pull/850

Issue #841, opened 2025-04-28
Handle deleted images


frzsombor:
I think there is an issue to be fixed, when a user removes an image from the Media Library and re-uploads a modified version, with the same name (to the same "month" folder).

The issue is that when deleting an image, neither the plugin, nor WordPress remove the optimized files and the backups. Because of this, when the user finds some issues with the uploaded image, removes it, and uploads a new version, the plugin will think the image is already optimized, and even after a manual "Optimization Request", the optimized versions will remain the same (optimized images of the OLD version).

I temporarily fixed the issue by hooking `delete_attachment` action and removing the optimized images based on LSCache naming conventions:

```php
public function remove_plugin_replated_images( $post_id, $post ) {
	$image_path = wp_get_original_image_path( $post_id );
	$possible_paths = array();

	// Bail out if no image path
	// Unexpected edge case when LSCache triggers a cache purge (eg: during theme/plugin updates)
	if ( ! $image_path ) {
		return;
	}

	// Optimized versions of original
	$possible_paths[] = $image_path . '.avif';
	$possible_paths[] = $image_path . '.webp';

	// Optimized versions of thumnails
	// Idea: https://github.com/WordPress/wordpress-develop/blob/6.8/src/wp-includes/post.php#L6729-L6743
	$upload_path = trailingslashit( dirname( $image_path ) );
	$meta = wp_get_attachment_metadata( $post_id );

	if ( isset( $meta['sizes'] ) && is_array( $meta['sizes'] ) ) {
		foreach ( $meta['sizes'] as $size => $sizeinfo ) {
			if ( $sizeinfo['file'] ) {
				$extension = pathinfo( $sizeinfo['file'], PATHINFO_EXTENSION );
				$file_without_ext = substr( $sizeinfo['file'], 0, -strlen( $extension ) );
				$possible_paths[] = $upload_path . $sizeinfo['file'] . '.avif';
				$possible_paths[] = $upload_path . $sizeinfo['file'] . '.webp';
				$possible_paths[] = $upload_path . $file_without_ext . 'bk.' . $extension;
				$possible_paths[] = $upload_path . $file_without_ext . 'bk.optm.' . $extension;
			}
		}
	}

	// Backups
	// Source: https://github.com/litespeedtech/lscache_wp/blob/v7.1/src/media.cls.php#L298-L300
	$extension = pathinfo( $image_path, PATHINFO_EXTENSION );
[...]

PR #844, opened 2025-05-08
Fix Media delete

timotei-litespeed:
Media delete is not working when run from AJAX.

### How to reproduce:
Go to Media
Switch style to grid
Upload a image
Run optimization
Go back to media
Choose a image
Delete from popup(bottom right, click on **Delete permanently**)

### Fix
This PR will delete the images when added from AJAX request

### DIFF
--- a/src/core.cls.php
+++ b/src/core.cls.php
@@ -178,6 +178,9 @@ class Core extends Root
 			$this->cls('Admin');
 		}
 
+		// Hook to attachment delete action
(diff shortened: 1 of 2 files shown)


hi-hai:
This is to fix https://github.com/litespeedtech/lscache_wp/issues/841. Good catch! However, its not a good practice to run biz code directly in `core.cls`. I will give an amendment later.

hi-hai:
https://github.com/litespeedtech/lscache_wp/commit/52a796b4c67c0888d59f2d14b860df7afc2809cc

PR #848, opened 2025-05-09
Admin - "Disable all features" show on all pages

timotei-litespeed:
When "Disable all features" is active it will show the warning on all pages


hi-hai:
It shouldn't be put into admin init, otherwise it will show even its not a display visit but to run some admin codes.

Can just put above L546.

hi-hai:
@timotei-litespeed

PR #849, opened 2025-05-12
Admin - add callback for Click to copy elements

timotei-litespeed:
Elements: `Report ID` and `Copy Logs` now have callback to show that action is done.


tpl/toolbox/log_viewer.tpl.php
```diff
@@ -48,9 +48,25 @@ function subnav_link($item)
 		: '';
 	$label = isset($item['label']) ? $item['label'] : $item['name'];
 
+	if(isset($item['cssClass'])) {
+		$class = $class . ' ' . $item['cssClass'];
```
hi-hai, line 52: @timotei-litespeed `$class .=`

PR #850, opened 2025-05-12
Improvement change - use SQL DESCRIBE instead of SHOW TABLES LIKE

timotei-litespeed:
Thank you @boonebgorges  for PR: https://github.com/litespeedtech/lscache_wp/pull/834

Changed usage of `SHOW TABLES LIKE` to `DESCRIBE`
Added changes to save `suppress_errors` state and restore it.

### DIFF
--- a/src/data.upgrade.func.php
+++ b/src/data.upgrade.func.php
@@ -25,7 +25,11 @@ function litespeed_update_7_0_1()
 	Debug2::debug('[Data] v7.0.1 upgrade started');
 
 	$tb_url = $wpdb->prefix . 'litespeed_url';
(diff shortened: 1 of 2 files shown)


src/data.upgrade.func.php
```diff
@@ -106,7 +110,12 @@ function litespeed_update_5_3()
 {
 	global $wpdb;
 	Debug2::debug('[Data] Upgrade url_file table');
-	$tb_exists = $wpdb->get_var('SHOW TABLES LIKE "' . $wpdb->prefix . 'litespeed_url_file"');
+
+	$save_state = $wpdb->suppress_errors;
```
hi-hai, line 114: @timotei-litespeed this duplicate lines could be a func. I will update it. Just remind you.

PR #875, opened 2025-07-03
Small changes

timotei-litespeed:


hi-hai:
@timotei-litespeed ` WARNING | [ ] json_encode() is discouraged. Use wp_json_encode() instead.` There are also other warnings. We can leave the files as long as no checker warning.

PR #885, opened 2025-07-31
Admin - show filter/server/constant value for settings that have them applied

timotei-litespeed:
Add admin overwrite to use filter value
Add admin overwrite to use SERVER value
Change display for text area settings

### DIFF
--- a/src/admin-display.cls.php
+++ b/src/admin-display.cls.php
@@ -46,6 +46,65 @@ class Admin_Display extends Base {
 
 	private $_btn_i = 0;
 
+	/*
+	 * List of settings with filters and return type.
+	 * 
+	 * @since  7.4
+	 */
+	protected static $settings_filters  = array(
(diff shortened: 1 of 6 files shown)


src/admin-display.cls.php
```diff
@@ -46,6 +46,65 @@ class Admin_Display extends Base {
...
 	private $_btn_i = 0;
 
+	/*
+	 * List of settings with filters and return type.
+	 * 
+	 * @since  7.4
+	 */
+	protected static $settings_filters  = array(
+		// Crawler - Blocklist
+		'crawler-blocklist' => [
```
hi-hai, line 56: keep in mind short array tag will not pass format checker

hi-hai: @timotei-litespeed Actually short tag is easier to read and doesn't hurt. I will update the rule to allow it.

PR #886, opened 2025-07-31
Disable all functionalities for 24h

timotei-litespeed:
Disable all functionalities for 24H
Added warning in Beta Test that LSC is disabled

### DIFF
--- a/src/debug2.cls.php
+++ b/src/debug2.cls.php
@@ -49,6 +49,51 @@ class Debug2 extends Root {
 	}
 
 	/**
+	 * Disable all functionalities for a time period.
+	 *
+	 * @since 7.4
+	 * @access public
+	 * @param  integer $time How long should we disable LSC functionalities.
+	 */
+	public static function temporary_disable( $time = 86400 ) {
+		$conf = Conf::cls();
+		$disabled = self::cls()->conf( Base::DEBUG_TMP_DISABLE );
+
+		if ( 0 === $disabled ) {
+			$conf->update_confs( [ Base::DEBUG_TMP_DISABLE => time() + $time ] );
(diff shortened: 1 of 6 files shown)


src/debug2.cls.php
```diff
@@ -48,6 +48,51 @@ public function __construct() {
...
+	 *
+	 * @since 7.4
+	 * @access public
+	 * @param  integer $time How long should we disable LSC functionalities.
+	 */
+	public static function temporary_disable( $time = 86400 ) {
+		$conf = Conf::cls();
+		$disabled = self::cls()->conf( Base::DEBUG_TMP_DISABLE );
+
+		if ( 0 === $disabled ) {
```
hi-hai, line 62: @timotei-litespeed If it is still in tmp disable, should still allow a refresh to extend to another 24h

src/debug2.cls.php
```diff
@@ -48,6 +48,51 @@ public function __construct() {
...
+			$conf->update_confs( [ Base::DEBUG_TMP_DISABLE => time() + $time ] );
```
hi-hai, line 63: Short array tag is not allowed by the format checker

src/debug2.cls.php
```diff
@@ -48,6 +48,51 @@ public function __construct() {
...
+	 * @access public
+	 */
+	public static function is_temporary_disable() {
+		$conf = Conf::cls();
+		$disabled_time = self::cls()->conf( Base::DEBUG_TMP_DISABLE );
+
+		if ( 0 === $disabled_time ) {
+			return false;
+		} else {
+			$time =  time();
```
hi-hai, line 86: Redundant var

PR #919, opened 2025-10-14
Dark Mode - apply faster

timotei-litespeed:


hi-hai:
That is strange. https://github.com/litespeedtech/lscache_wp/blob/dev/src/admin-display.cls.php#L423 its already true @timotei-litespeed

PR #972, opened 2026-04-22
Auto purge Cloudflare cache on post/page updates when Cloudflare API is enabled

AsFLY-902:
- Added automatic Cloudflare cache purge on post/page updates when the Cloudflare API option is enabled.
- Ensures that updated content is immediately reflected without requiring a full cache purge.


hi-hai:
Per page purge will consume CF quota quickly. Besides, we have purge CF in purge all already. If add purge 3rd party feature, will need to use hook other than call CF directly in Purge class. I will close this PR for now. In future if we need to add similar feature, will design ATM.

Still a good try! Thanks

PR #976, opened 2026-05-07
Fix LQIP generation failing for native WebP/AVIF images

AsFLY-902:
## Problem

LQIP generation fails for natively uploaded WebP and AVIF images. The `Utility::drop_webp()` function unconditionally strips `.webp`/`.avif` extensions, which breaks URLs for native WebP/AVIF files that don't have a double extension (e.g., `image.webp` vs `image.png.webp`).

**Example:**
- `image.png.webp` → `image.png` Success (LiteSpeed optimized, has source extension)
- `image.webp` → `image`  Fails (native WebP, results in 404)

This causes the LQIP service to either:
1. Get a 404 response and add the image to the exclude list 
2. Receive an HTML error page instead of image data (`Unsupported response content-type text/html`) and add the image to the exclude list 

## Fix

Refactored `Utility::drop_webp()` to only strip the `.webp`/`.avif` suffix when a valid source extension exists underneath (double-extension pattern like `.png.webp`). Native WebP/AVIF URLs are now returned unchanged.

## Changes

- `src/utility.cls.php` — Updated `drop_webp()` to use `pathinfo()` for proper extension detection instead of blind `substr()` stripping.

### DIFF
--- a/src/utility.cls.php
+++ b/src/utility.cls.php
@@ -446,10 +446,26 @@ class Utility extends Root {
 	 * @return string Cleaned filename.
 	 */
 	public static function drop_webp( $filename ) {
-		if ( in_array( substr( $filename, -5 ), [ '.webp', '.avif' ], true ) ) {
-			$filename = substr( $filename, 0, -5 );
+		$path = (string) wp_parse_url( $filename, PHP_URL_PATH );
+		if ( ! $path ) {
+			$path = $filename;
 		}
 
+		$base_ext = strtolower( pathinfo( $path, PATHINFO_EXTENSION ) );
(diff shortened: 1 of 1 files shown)


hi-hai:
@timotei-litespeed @AsFLY-902 Please give the latest dev a test to make sure this improved version works as expected.

timotei-litespeed:
@hi-hai from my tests the change is working as expected.

One thing I want to notice(not error, images are added to preload ): in page VPI metabox there are differences.
V7.8.1:
```
1
5188350___-1024x1024.jpg
```

V7.9Beta:
```
1.webp
5188350___-1024x1024.jpg
```
Notice the **.webp** added to first image in 7.9

PR #977, opened 2026-05-13
Thirdparty - add plugin RankMath SEO

timotei-litespeed:


thirdparty/rank-math.cls.php
```diff
@@ -0,0 +1,65 @@
...
+ */
+class Rank_Math {
+
+	/**
+	 * Detects if Rank Math is active.
+	 *
+	 * @since 7.9
+	 * @return void
+	 */
+	public static function detect() {
```
hi-hai: @timotei-litespeed What is this func for?

timotei-litespeed: @hi-hai the function is automatically run in **entry.inc.php**

hi-hai: Why?

PR #980, opened 2026-05-16
7.8.1 Refari - CDN defaults overhaul, plugin-wide settings UX, Object Cache Auto

RyansGit:
## CDN

- **Exclude Path presets** — new `data/cdn_exc.txt` ships with a bfcache-incompatible exclude (specifically `nocache-bfcache`'s `bfcache-invalidation.js`, which must be served from origin so the back/forward cache invalidation shim isn't cached/rewritten by the CDN). New `litespeed_cdn_exc` filter lets devs extend or override the list. 
- **Include File Types** default trimmed from 19 → 3: now `.css`, `.less`, `.js`. The 16 dropped extensions (`.aac .eot .gif .jpeg .jpg .mp3 .mp4 .ogg .otf .pdf .png .svg .ttf .webp .woff .woff2`) are now owned by the dedicated *Include Images / Docs / Fonts / Media* toggles — closes the long-standing footgun where toggling *Include Fonts* OFF still let fonts through via the filetype textarea.
- **Auto-append** `.css` / `.less` / `.js` to the filetype list when the corresponding CSS/JS toggle is switched ON.
- **"Default value" panel** rendered next to *Include File Types*, with inline `missing` / `inconsistent` chips so users see exactly which extensions deviate from the bundled default.
- **New CDN Mapping toggles**: *Include Docs*, *Include Fonts*, *Include Media* (alongside the existing Images / CSS / JS).

## Admin UI — Tabs & Switches

- **Tab-flash fix** carried from the Cache settings page to every tabbed entry template (CDN, Crawler, DB Optm, General, Image Optm, Page Optm, Presets, Toolbox). Each layout panel now ships with an `id` matching its tab key plus `data-litespeed-default-tab="1"` on the cookie-selected (or first) tab, so the correct panel paints before JS runs.
- **Flash-prevention CSS consolidated** into `litespeed.css` instead of duplicated inline per template.
- **Admin JS** suppresses the default anchor-jump on tab click and updates the URL fragment via `history.replaceState` so the tab stays shareable for the no-JS fallback path.
- **Sliding-pill toggle switches** — animated, with hover marker to improve UX.
- **`--litespeed-switch-count`** CSS var emitted on every server-rendered switch so the indicator pill lines up correctly.
- **Switch labels** forced to a single line (Debug Admin IP)
- **Dark mode** preference mirrored to a cookie so PHP can render the correct `<body>` class on first paint (no FOUC).
- **Style refactor** across entry templates: inlined the temporary default-tab variable into the echoed div and switched object-cache `kind` comparisons to Yoda (constant-first) form.

## Admin UI — Defaults / Recommendations

- **Scalar defaults** — empty scalar settings now render a `Default value: <code>X</code>` hint, a missing-value flag, and a **"+ Use default value"** link that writes the bundled default into the field on click. Auto-emitted from `build_textarea()`; templates only need `skip_recommended=true` when they wrap their own side-by-side layout. Emission tracked per option ID to prevent duplicate panels.
- **Array defaults** — `recommended()` helper now compares the cached default list against the saved value (case-insensitive, trimmed) and emits a `litespeed-defaults-flags` chip row plus a discrete **"+ Add missing defaults (N)"** text link. A single delegated jQuery handler appends missing values without case-insensitive duplication and hides the chip row + link once applied. Every page using `recommended()` (HTML Attribute To Replace, Included Directories, Heartbeat TTLs, Media Lazy Exc, Localization, Cache TTL, Crawler, etc.) gets this affordance for free — no template changes needed.
- **Admin JS click handler** extended with a new `set` mode that replaces the target field's value outright (works for both `<textarea>` and `<input>`), in addition to the prior "add missing" mode.

## Object Cache

- **Auto method** added — picks the best available backend on save instead of forcing admins to know whether the host has Redis/Valkey, Memcached, a TCP listener or a Unix socket. **Now the default for fresh installs.**
- Detection priority (Redis/Valkey before Memcached):
  1. Unix sockets that exist on disk (cPanel / LSWS convention first)
  2. The configured Host/Port if the admin typed one
  3. `localhost` on the default port
  4. `127.0.0.1` on the default port
- On Save the persisted kind/host/port are rewritten to the detected backend, so the *Auto* sentinel never reaches runtime. If Object Cache is off, *Auto* coerces back to Memcached.
- **Status panel** rewritten to show *UNIX Socket / Redis-Valkey / Memcached / Connection Test* in that order, with a *Passed (auto-detected)* state. Connection-test transient busted on save so the panel reflects new settings immediately.
- **Copy tweaks**: LSMCD unlinked in Host description (Deprecated), Redis/Valkey link added in the Host description, socket path reworded as an example, Method help text simplified.

## Plumbing / Internals

- **DB upgrader 7.8.1.1-b1** (`litespeed_update_7_8_1_1_basic_placeholder`) — backfills the bundled gray base64 GIF into `media-lazy_placeholder` for sites with Lazy Load enabled and an empty placeholder. Manually-blanked values are preserved on subsequent loads via the versioned upgrader. @timotei-litespeed @hi-hai this might need a different solution
- **`media-auto_rescale_ori`** option registered in `data/const.default.json`.
- **Stored default for `media-lazy_placeholder`** seeded.
- **Translation strings** added (`src/lang.cls.php`).
- **CLI option wiring** (`cli/option.cls.php`).
- Supporting changes across `src/base.cls.php`, `src/data.cls.php`, `src/cdn.cls.php`, `src/admin-display.cls.php`, `src/admin-settings.cls.php`.

If you need a loom video to demonstrate the changes for easier checking, drop me a slack message :) (Ryan D)


hi-hai:
@RyansGit Thanks Ryan. Do you mind to split the PRs to separate ones per feature update (one thing at a time)? That would make review and management easier. Also, please use `dev` branch as target.

PR #984, opened 2026-05-19
fix(admin): stop the wrong tab from flashing before page settles

RyansGit:
When opening any LiteSpeed admin page with tabs (Cache, CDN, Crawler, etc.) the first tab briefly showed before jumping to the user's last-selected tab. The correct tab is now picked server-side from the remembered cookie, so the right panel paints straight away. Tab clicks also keep the URL fragment shareable without scrolling the page.

### DIFF
--- a/assets/css/litespeed.css
+++ b/assets/css/litespeed.css
@@ -527,6 +527,12 @@ button.litespeed-form-action:hover {
 	padding: 1px 20px 20px 20px;
 }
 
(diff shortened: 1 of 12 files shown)


hi-hai:
Done in https://github.com/litespeedtech/lscache_wp/commit/3494b61cb11d069b97156c4e3ff38a169ef18d46

PR #998, opened 2026-06-05
OC - validate host

timotei-litespeed:
Validate OC host to be valid: IP, host name or absolute path in case of socket usage.
Works with address: `::1`

Ticket: [https://wordpress.org/support/topic/possible-bug-with-in-redis-path-in-litespeed-cache-wp-plugin/#post-18929290](https://wordpress.org/support/topic/possible-bug-with-in-redis-path-in-litespeed-cache-wp-plugin/#post-18929290)


hi-hai:
Merged. Did some minor tuning. Thx

PR #1002, opened 2026-06-11
fix(optm): css combine now matches placeholder link tag whitespace

### LINKED ISSUE #999: Page Optimize: CSS Combine emits an unstyled page
## Summary
 
With **CSS Combine** enabled, a page can be served with **no CSS at all** — every original `<link>` stripped, the combined `<link>` never injected, and the `litespeed-dummy.css` placeholder left in the output. The combined CSS file is generated on disk correctly; only its injection into the HTML fails. Fully reproducible and survives every regeneration, so a cache purge does not help — it just re-caches another unstyled page.
 
**Affected versions:** 7.8, 7.8.0.1, 7.8.1 (latest), and `master`/trunk — the relevant code is identical in all.
 
### 1) `DUMMY_CSS_REGEX` only tolerates a single space between the `id` and `href` attributes
 
```php
const DUMMY_CSS_REGEX = "#<link [ \w='\"/]*id=['\"]litespeed-cache-dummy-css['\"] href=['\"].+assets/css/litespeed-dummy\.css[?\w.=-]*['\"][ \w='\"/]*>#isU";
```
 
The flexible `[ \w='\"/]*` groups sit before `id=` and after the `href` value, but between the `id` attribute and `href` there is a **single literal space**. If anything renders the placeholder tag with two spaces (a common result of `style_loader_tag` filters / markup normalisation by other plugins), the pattern no longer matches. The v7.2 changelog ("better compatibility for dummy CSS removal in cases where other plugins manipulate the quotation marks") hardened the quotes but not the whitespace.
 
Runnable proof:
 
```php
$re  = "#<link [ \w='\"/]*id=['\"]litespeed-cache-dummy-css['\"] href=['\"].+assets/css/litespeed-dummy\.css[?\w.=-]*['\"][ \w='\"/]*>#isU";
$one = "<link rel='stylesheet' id='litespeed-cache-dummy-css' href='https://example.com/wp-content/plugins/litespeed-cache/assets/css/litespeed-dummy.css' type='text/css' media='all' />";
$two = "<link rel='stylesheet' id='litespeed-cache-dummy-css'  href='https://example.com/wp-content/plugins/litespeed-cache/assets/css/litespeed-dummy.css' type='text/css' media='all' />";
// ^ two spaces before href
 
var_dump(preg_match($re, $one)); // int(1)
var_dump(preg_match($re, $two)); // int(0)  <-- extra whitespace => no match
```
 
### 2) The injection guard treats "no match" as success, silently dropping the optimized CSS instead of falling back
 
The head-injection site is guarded with `false !== preg_match(...)`:
 
```php
// Put header content to dummy css position
if (false !== preg_match(self::DUMMY_CSS_REGEX, $this->content)) {
    self::debug('Put optm data to dummy css location');
    $this->content = preg_replace(self::DUMMY_CSS_REGEX, $this->html_head, $this->content);
}
elseif ( /* <meta charset> present */ ) {
    self::debug('Put optm data to be after <meta charset>');
    // ...inject $this->html_head after <meta charset>
}
else {
    self::debug('Put optm data to be after <head>');
    // ...inject $this->html_head after <head>
}
```
 
`preg_match()` returns `0` on no-match and `false` only on error, so `false !== preg_match(...)` is true **even when the pattern does not match**. On the 2-space tag, `preg_match` returns `0`, the code wrongly enters the "matched" branch, runs a `preg_replace` that changes nothing, and never reaches the `<meta charset>` / `<head>` fallback — so `$this->html_head` (which contains the combined CSS `<link>`) is discarded entirely.
 
The "drop dummy" fallback has the same guard and so also no-ops, which is why the placeholder is additionally left behind in the output:
 
```php
// Fallback to replace dummy css placeholder
if (false !== preg_match(self::DUMMY_CSS_REGEX, $content)) {
    self::debug('Fallback to drop dummy CSS');
    $content = preg_replace(self::DUMMY_CSS_REGEX, '', $content);
}
```
 
Debug-log fingerprint (every generation of an affected page):
 
```
🎢  Put optm data to dummy css location
🎢  Fallback to drop dummy CSS
```
 
## Steps to reproduce
 
1. **Page Optimization → CSS Settings:** enable **CSS Combine** (with Minify / Combine External & Inline).
2. On a site where some `style_loader_tag` filter (or markup) renders the enqueued `litespeed-cache-dummy-css` `<link>` with **two spaces** between `id='…'` and `href` (verify in the raw, unoptimized HTML).
3. Load any cacheable front-end page.
**Result:** the served HTML contains the un-replaced `litespeed-dummy.css` placeholder, no `…/wp-content/litespeed/css/<hash>.css` link, and none of the original stylesheets → unstyled page. The combined file exists under `wp-content/litespeed/css/` but is never referenced.
 
## Suggested fix
 
- Correct the guards from `false !== preg_match(...)` to `1 === preg_match(...)` so a non-match correctly falls through to the `<meta charset>` / `<head>` insertion. This alone prevents the unstyled-page failure regardless of the regex.
- Additionally, make `DUMMY_CSS_REGEX` whitespace-tolerant (e.g. ` +href=` / `\s+` instead of the single literal space) so the placeholder is matched and replaced in place.
## Environment
 
- LiteSpeed Cache 7.8.1 (also confirmed in `master`)
- WordPress 6.x
- PHP 7.4
- LiteSpeed Web Server
Combine generation works; only HTML injection fails.

Image

faisalahammad:
## Summary

CSS Combine could serve a fully unstyled page when the placeholder link tag used anything other than a single literal space between the `id` and `href` attributes (which is the default WordPress render for `wp_print_styles`). The bug came from two compounding defects in `src/optimize.cls.php`: a too-strict regex and a guard that treated a no-match as success.

Fixes #999

### `src/optimize.cls.php` — `DUMMY_CSS_REGEX`

**Before:**
```php
const DUMMY_CSS_REGEX = "#<link [ \w='\"/]*id=['\"]litespeed-cache-dummy-css['\"] href=['\"].+assets/css/litespeed-dummy\.css[?\w.=-]*['\"][ \w='\"/]*>#isU";
```

**After:**
```php
const DUMMY_CSS_REGEX = "#<link [ \w='\"/]*id=['\"]litespeed-cache-dummy-css['\"][ \w='\"/]*href=['\"].+assets/css/litespeed-dummy\.css[?\w.=-]*['\"][ \w='\"/]*>#isU";
```

**Why:** the literal single space between the id value and `href=` is the gap that fails on WordPress's default two-space padding. Reusing the existing `[ \w='\"/]*` character class keeps the pattern consistent and matches one, two, or more spaces (matches the pattern both maintainer and reporter agreed on in the issue thread).

### `src/optimize.cls.php` — both `preg_match` guards

**Before:**
```php
// site 1: line 503
if (false !== preg_match(self::DUMMY_CSS_REGEX, $this->content)) {

// site 2: line 228
if (false !== preg_match(self::DUMMY_CSS_REGEX, $content)) {
```

**After:**
```php
if (1 === preg_match(self::DUMMY_CSS_REGEX, $this->content)) {
if (1 === preg_match(self::DUMMY_CSS_REGEX, $content)) {
```

**Why:** `preg_match()` returns `0` on no-match and `false` only on error. The old `false !==` check was true on `0`, so when the regex failed to match the code entered the "matched" branch, ran a no-op `preg_replace`, and the `elseif`/`else` fallbacks (meta charset / head insertion) never ran. The combined CSS was silently dropped. `1 ===` makes no-match correctly fall through. This is the real safety net for any future markup variation that fails the regex.

### `readme.txt`

Added a `* **Page Optimize**` bullet to the 7.8.1 changelog referencing #999.

## Testing

**Test 1: original repro on stock WP 6.0.x**
1. Enable **Page Optimization → CSS Combine**.
2. Confirm the placeholder link tag has two spaces between `id='…'` and `href='…'` in the raw HTML.
3. Load a cacheable front-end page and view source.
4. Result: combined CSS link present, placeholder gone, page styled.

**Test 2: whitespace variants**
1. With CSS Combine enabled, load pages on a stock install.
2. Result: placeholder replaced regardless of one, two, or more spaces between `id` and `href`.

**Test 3: guard fallback (synthetic non-match)**
1. Inject a `data-*` attribute between `id` and `href` via a `style_loader_tag` filter (edge case the char class still does not cover).
2. With CSS Combine enabled, purge cache, load a page.
3. Result: combined CSS injected after `<meta charset>` (debug log "Put optm data to be after <meta charset>") instead of being silently dropped.

**Test 4: PHPCS**
1. `composer sniff-check` and `npm run format-check` pass on changed files. The only failure in the repo is pre-existing and unrelated, in `src/cdn/cloudflare.cls.php` lines 249 and 277.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -259,6 +259,7 @@ You can report security bugs through the Patchstack Vulnerability Disclosure Pro
 
 = 7.8.1 - Apr 1 2026 =
 * **CDN** Fixed Cloudflare API key type detection for the compatibility w/ the new key format.
+* **Page Optimize** Fixed CSS Combine serving unstyled pages when the placeholder link tag uses default WordPress whitespace. (#999)
 
 = 7.8.0.1 - Mar 17 2026 =
 * **Object Cache** Improved Object Cache resilience: auto disable when connection fails, network subsites fallback to database, and dropped TTL setting to respect never-expired transients.
--- a/src/optimize.cls.php
+++ b/src/optimize.cls.php
@@ -20,7 +20,7 @@ class Optimize extends Base {
 
 	const ITEM_TIMESTAMP_PURGE_CSS = 'timestamp_purge_css';
 
-	const DUMMY_CSS_REGEX = "#<link [ \w='\"/]*id=['\"]litespeed-cache-dummy-css['\"] href=['\"].+assets/css/litespeed-dummy\.css[?\w.=-]*['\"][ \w='\"/]*>#isU";
+	const DUMMY_CSS_REGEX = "#<link [ \w='\"/]*id=['\"]litespeed-cache-dummy-css['\"][ \w='\"/]*href=['\"].+assets/css/litespeed-dummy\.css[?\w.=-]*['\"][ \w='\"/]*>#isU";
 
 	private $content;
 	private $content_ori;
@@ -225,7 +225,7 @@ class Optimize extends Base {
 	public function finalize( $content ) {
 		$content = $this->_finalize($content);
 		// Fallback to replace dummy css placeholder
-		if (false !== preg_match(self::DUMMY_CSS_REGEX, $content)) {
+		if (1 === preg_match(self::DUMMY_CSS_REGEX, $content)) {
 			self::debug('Fallback to drop dummy CSS');
 			$content = preg_replace( self::DUMMY_CSS_REGEX, '', $content );
 		}
@@ -500,7 +500,7 @@ class Optimize extends Base {
 				$this->content = str_replace('</head>', $this->html_head . '</head>', $this->content);
 			} else {
 				// Put header content to dummy css position
-				if (false !== preg_match(self::DUMMY_CSS_REGEX, $this->content)) {
+				if (1 === preg_match(self::DUMMY_CSS_REGEX, $this->content)) {
 					self::debug('Put optm data to dummy css location');
 					$this->content = preg_replace( self::DUMMY_CSS_REGEX, $this->html_head, $this->content );
 				}


hi-hai:
https://github.com/litespeedtech/lscache_wp/commit/7dc6a0117a978cb6f38231dbb3877f1acea3007d

PR #1006, opened 2026-06-23
Fix ESI nonce generation for actions with custom nonce lifetimes.

robertstaddon:
WordPress 6.1 allows wp_nonce_tick() to receive the nonce action so plugins can customize nonce lifetime per action via nonce_life. LiteSpeed’s ESI nonce generator was calling wp_nonce_tick() without the action, so ESI-generated nonces could differ from the values later expected by wp_verify_nonce() / check_ajax_referer(). This broke plugins such as Gravity Forms, which sets a custom 3-day lifetime for gform_ajax_submission.

This change makes ESI nonce generation action-aware where supported and sets the ESI nonce fragment TTL based on the effective action-specific nonce tick length.

### DIFF
--- a/litespeed-cache.php
+++ b/litespeed-cache.php
@@ -184,9 +184,35 @@ if ( ! function_exists( 'litespeed_define_nonce_func' ) ) {
 			}
 
 			$token = wp_get_session_token();
-			$i     = wp_nonce_tick();
+			$i     = wp_nonce_tick_litespeed_esi( $action );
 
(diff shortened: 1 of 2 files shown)


hi-hai:
Looks good. Do you mind to update the target branch from `master` to `dev`? That is the branch we used for PRs.

robertstaddon:
Done!

timotei-litespeed:
@robertstaddon thank you for the changes
They have been merged and will appear in 7.9

PR #1024, opened 2026-07-28
Fix - Noscript removal in Optimize class

timotei-litespeed:


timotei-litespeed:
Not how plugin should work

PR #1027, opened 2026-08-06
Update esi.nonces.txt

kylewhenderson:
Predefined ESI nonce for Gravity Forms never matches — `gfrom_*` should be `gform_*`


timotei-litespeed:
@kylewhenderson Thank you. Pushed to our dev branch, and will be released with the new version

PR #1028, opened 2026-08-07
fix(object-cache): ensure wp_cache_delete returns boolean

### LINKED ISSUE #924: Type error: wp_cache_delete can return null
Due to this method returning bool|null:
https://github.com/litespeedtech/lscache_wp/blob/ae7624ebd11fd4b832987a7c1a0779e5488f078c/src/object-cache.cls.php#L740-L749

wp_cache_delete returns bool|null even though it's documented as returning bool:
https://github.com/litespeedtech/lscache_wp/blob/ae7624ebd11fd4b832987a7c1a0779e5488f078c/src/object.lib.php#L202-L212

faisalahammad:
## Summary
Guarantees that `wp_cache_delete()` and `WP_Object_Cache::delete()` strictly return a boolean (`true` or `false`) rather than `null`. This prevents PHP `TypeError` fatal exceptions in PHP 8.x when standard object cache functions are called.

Fixes #924

### Object Cache (`src/object.lib.php` & `src/object-cache-wp.cls.php`)
**Before:**
```php
function wp_cache_delete( $key, $group = '' ) {
	global $wp_object_cache;

	return $wp_object_cache->delete( $key, $group );
}
```
**After:**
```php
function wp_cache_delete( $key, $group = '' ) {
	global $wp_object_cache;

	if ( ! is_object( $wp_object_cache ) ) {
		return false;
	}

	return (bool) $wp_object_cache->delete( $key, $group );
}
```
**Why:** Directly returning the object cache method call could return `null` if the instance was uninitialized or returned non-boolean values. Adding an `is_object()` check and explicit `(bool)` typecast guarantees a boolean return value.

### WP Object Cache Class (`src/object-cache-wp.cls.php`)
**Before:**
```php
if ( $this->_object_cache->is_non_persistent( $group ) ) {
	return false;
}

return $this->_object_cache->delete( $id );
```
**After:**
```php
if ( $this->is_non_persistent( $group ) || ! is_object( $this->_object_cache ) ) {
	return $found;
}

return (bool) $this->_object_cache->delete( $id );
```
**Why:** Ensures deletion from runtime memory is returned for non-persistent groups or missing instances, and persistent store deletion output is explicitly cast to `bool`.

## Testing
**Test 1: Delete cache key**
1. Run `wp_cache_delete( 'test_key', 'default' )`.
2. Confirm return type is boolean `false` (or `true` if item exists) and never `null`.

**Test 2: Delete cache key with uninitialized object cache**
1. Set `$wp_object_cache = null` and invoke `wp_cache_delete( 'test_key' )`.
2. Confirm return value is `false` without throwing PHP `TypeError`.

### DIFF
--- a/src/object-cache-wp.cls.php
+++ b/src/object-cache-wp.cls.php
@@ -119,6 +119,15 @@ class WP_Object_Cache {
 	protected $global_groups = [];
 
 	/**
+	 * Non-persistent cache groups
+	 *
+	 * @since 1.8
+	 * @access protected
+	 * @var array
+	 */
+	protected $non_persistent_groups = [];
+
+	/**
 	 * Blog prefix for cache keys
 	 *
 	 * @since 1.8
@@ -271,7 +280,7 @@ class WP_Object_Cache {
 			$group = 'default';
 		}
 
-		$prefix = $this->_object_cache->is_global( $group ) ? '' : $this->blog_prefix;
+		$prefix = $this->is_global( $group ) ? '' : $this->blog_prefix;
 
 		return LSOC_PREFIX . $prefix . $group . '.' . $key;
 	}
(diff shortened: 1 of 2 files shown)


hi-hai:
Please use `dev` branch as the base

faisalahammad:
Rebased branch onto dev and updated target base to dev branch.

faisalahammad:
Resolved conflicts. PR is now clean against dev branch.

PR #1030, opened 2026-08-07
fix(media): prevent getimagesize loopback deadlocks on missing images

### LINKED ISSUE #894: [BUG] getimagesize() on missing image causes LSAPI deadlock and full site outage (PHP 8.3.24, LSCache 7.4)
Environment

PHP version: 8.3.24

LiteSpeed Cache for WordPress: 7.4

Web Server: OpenLiteSpeed (OLS)

We're experiencing critical issues where the site goes down due to LSAPI deadlocks caused by the LiteSpeed Cache plugin.

When the plugin attempts to call getimagesize() on a missing image (typically under /wp-content/uploads/), the following error appears repeatedly in the logs:

`[STDERR] PHP Warning:  getimagesize(https://example.com/wp-content/uploads/2021/10/example_file.png): Failed to open stream: HTTP request failed! in /folder/example.com/wp-content/plugins/litespeed-cache/src/media.cls.php on line 899
`

This is immediately followed by OpenLiteSpeed logs like:

`No request delivery notification has been received from LSAPI application, possible dead lock`

This stalls PHP workers, causes ExtConn timed out while processing messages, and ultimately results in the entire WordPress site becoming unresponsive.

Root Cause
It appears that getimagesize() is used without a prior check to confirm the existence or availability of the image URL. When the image is missing or the HTTP call fails, the process hangs and blocks the PHP worker without timing out gracefully.

Expected Behavior
The plugin should verify that the image exists before attempting to process or cache it. Failing that, it should handle the error gracefully without crashing the PHP worker or causing LSAPI deadlocks.

Suggested Fix
Add a check before getimagesize():

```
if (@get_headers($url)[0] !== 'HTTP/1.1 404 Not Found') {
    $info = getimagesize($url);
}
```

Or at least wrap the getimagesize() call in a suppression and error-handling block to prevent hanging on 404s.

faisalahammad:
## Summary
Fix getimagesize() LSAPI deadlocks on missing image URLs. When image URLs belong to internal site or CDN host but the file is missing locally, prevent getimagesize() from making self-referencing HTTP loopback requests back to OpenLiteSpeed web server.

Fixes #894

### Media (`src/media.cls.php`)
**Before:**
```php
private function _detect_dimensions( $src ) {
    $pathinfo = Utility::is_internal_file( $src );
    if ( $pathinfo ) {
        $src = $pathinfo[0];
    } elseif ( apply_filters( 'litespeed_media_ignore_remote_missing_sizes', false ) ) {
        return false;
    }
    // ...
    $sizes = getimagesize( $src );
```
**After:**
```php
private function _detect_dimensions( $src ) {
    $pathinfo = Utility::is_internal_file( $src );
    if ( $pathinfo ) {
        $src = $pathinfo[0];
    } elseif ( Utility::is_internal_url( $src ) || apply_filters( 'litespeed_media_ignore_remote_missing_sizes', false ) ) {
        return false;
    }
    // ...
    $sizes = getimagesize( $src );
```
**Why:** If `is_internal_file()` returns `false` on an internal URL because the image is missing on disk, checking `Utility::is_internal_url( $src )` returns `false` early. This skips `getimagesize()` so PHP workers do not make HTTP loopback GET calls to the same web server.

### Utility (`src/utility.cls.php`)
**Before:**
```php
// Only is_internal_file existed, which requires file to exist on local disk
```
**After:**
```php
public static function is_internal_url( $url ) {
    if ( 'data:' === substr( $url, 0, 5 ) ) {
        return false;
    }

    $url_parsed = wp_parse_url( $url );
    if ( empty( $url_parsed['host'] ) ) {
        return true;
    }

    if ( self::internal( $url_parsed['host'] ) || CDN::internal( $url_parsed['host'] ) ) {
        return true;
    }

    return false;
}
```
**Why:** Checks if a URL host matches the internal site or internal CDN domain without checking file existence on disk.

## Testing

**Test 1: Missing Local Image URL**
1. Enable Add Missing Sizes in Media optimization settings.
2. Load page containing an image tag with a missing local uploads URL.
Result: Page loads immediately without delays or PHP getimagesize warnings in error log.

**Test 2: Valid Local Image**
1. Load page with an existing upload image without width/height attributes.
Result: Width and height dimensions are calculated and injected from local filesystem path as expected.

### DIFF
--- a/src/media.cls.php
+++ b/src/media.cls.php
@@ -1146,7 +1146,7 @@ class Media extends Root {
 		$pathinfo = Utility::is_internal_file( $src );
 		if ( $pathinfo ) {
 			$src = $pathinfo[0];
-		} elseif ( apply_filters( 'litespeed_media_ignore_remote_missing_sizes', false ) ) {
+		} elseif ( Utility::is_internal_url( $src ) || apply_filters( 'litespeed_media_ignore_remote_missing_sizes', false ) ) {
 			return false;
 		}
 
@@ -1156,7 +1156,7 @@ class Media extends Root {
 
 		try {
 			$sizes = getimagesize( $src );
-		} catch ( \Exception $e ) {
+		} catch ( \Throwable $e ) {
 			return false;
 		}
 
--- a/src/media.cls.php
+++ b/src/media.cls.php
@@ -1146,7 +1146,7 @@ class Media extends Root {
 		$pathinfo = Utility::is_internal_file( $src );
 		if ( $pathinfo ) {
 			$src = $pathinfo[0];
-		} elseif ( Utility::is_internal_url( $src ) || apply_filters( 'litespeed_media_ignore_remote_missing_sizes', false ) ) {
+		} elseif ( $this->_is_internal_url( $src ) || apply_filters( 'litespeed_media_ignore_remote_missing_sizes', false ) ) {
 			return false;
 		}
 
@@ -1165,6 +1165,27 @@ class Media extends Root {
 		}
 
 		return false;
+	}
+
+	/**
+	 * Check if a URL belongs to an internal site or CDN domain.
+	 *
+	 * @since 7.9
+	 *
+	 * @param string $url URL.
+	 * @return bool True if internal host or relative path.
+	 */
+	private function _is_internal_url( $url ) {
+		if ( 'data:' === substr( $url, 0, 5 ) ) {
+			return false;
+		}
+
+		$url_parsed = wp_parse_url( $url );
+		if ( empty( $url_parsed['host'] ) ) {
+			return true;
+		}
+
+		return Utility::internal( $url_parsed['host'] ) || CDN::internal( $url_parsed['host'] );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


hi-hai:
Please use `dev` branch as base.

timotei-litespeed:
@faisalahammad please make sure there is 1 bug fix per PR.
Thank you

faisalahammad:
Rebased on dev branch with single commit for issue #894 fix. Thank you!

timotei-litespeed:
@faisalahammad there are some conflicts to solve

faisalahammad:
I have rebased the branch cleanly on dev branch with 1 commit. The merge conflicts are resolved.

timotei-litespeed:
@faisalahammad thank you! :)
I will look over the change on Monday

faisalahammad:
Addressed reviewer feedback in b80438ca:\n\n- The deadlock fix is the internal site/CDN host check before getimagesize(); is_internal_file() alone returns false when the internal file is missing, which is the deadlock case.\n- Removed the new public Utility::is_internal_url() method. Host matching now reuses Utility::internal() and CDN::internal() through a private Media helper.\n- Retained the \\Throwable catch because PHP 8 can raise ValueError for an empty filename.\n\nManual testing confirmed. No inline review threads were available through the GitHub API, so this is posted as the PR-level response.

